PR #41, opened 2013-07-03
A plugin to avoid slug collisions for articles with the same title

leofiore:


jonathan-s:
this pull request could be closed as it's already implemented in core.

leofiore:
actually, looking at the code, i don't see anything similar, nor i see any relevant testcase.

PR #90, opened 2013-10-11
Lightbox Plugin: Basic Lightbox support for images and figures

jprine:
Lightbox works for image tags with a class of the form `lb-[set]` where `[set]` is the Lightbox image set the image will be placed in.

The Lightbox css, javascript, and image files still need to be added manually to the theme.


antgel:
This has been around for a while. Shame for those of us finding the pure CSS lightbox a bit vanilla.

justinmayer:
@antgel: I assume the submodule version by @kura didn't work well for you? If it did not, do you know whether the code in this pull request works well with the latest Pelican 4.0 release?

justinmayer:
See also: #824

antgel:
@justinmayer Indeed, the submodule you refer to (and which I meant by "pure CSS lightbox") requires thumbnails to be created separately to main images. The beauty of the `{lightbox}` notation in the Photos plugin is that it doesn't require such heavy lifting. It's generally confusing that the Photos plugin is designed to work with a different Lightbox than is supplied in this repository. The only (vague) clue is that in the Photo plugin docs, the link is to https://lokeshdhakar.com/projects/lightbox2/, but it's still not clear to the newbie.

antgel:
Doubtful, I think lightbox2 is the de facto for a while.

PR #92, opened 2013-10-12
Make liquid video support relative url path and pelican internal content.

calfzhou:
I'd like to make liquid video tag support relative url path (such as `../videos/xxx.mp4`) and pelican internal content (e.g. `{filename}../videos/2013/xxx.mp4`).

A unit test is also added to verify the functionality.

### DIFF
--- a/liquid_tags/test_video.py
+++ b/liquid_tags/test_video.py
@@ -1,0 +1,39 @@
+import unittest
+from .video import video
+
+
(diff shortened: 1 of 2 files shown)


justinmayer:
Hi @calfzhou. It took nearly seven years, but we finally got your contribution merged via 5cec402. I updated the tests in ab31dd6. 🎉

PR #111, opened 2013-11-17
Youtube align

vellvisher:
Added text-align attribute to liquid_tags.youtube by enclosing the iframe in an HTML <div> tag and specifying the text-align property.

Also, shifted to named groups for Python regex.

Thought someone might want to do the same thing!


kdeldycke:
The `liquid_tags` plugin has been migrated as a stand-alone plugin at https://github.com/pelican-plugins/liquid-tags . I recommend you to try out that new version as some bugs were fixed and the plugin is actively being worked on over there.

In the mean time I guess we can close this PR.

justinmayer:
Agreed. Thanks, Kevin!

PR #139, opened 2014-02-01
new plugin: i18n_subsites

smartass101:
This plugin provides a simple way to create subsites for each used language.
Themes can be localized by specifying a different _THEME_ for each language in _I18N_SUBSITES_
In the future theming should support `jinja2.ext.i18n` + babel too, but this first version of the plugin concentrates mainly on the generation aspect.

### DIFF
--- a/i18n_subsites/README.rst
+++ b/i18n_subsites/README.rst
@@ -1,11 +1,12 @@
-i18n subsites plugin
-===================
+======================
+ i18n subsites plugin
+======================
 
 This plugin extends the translations functionality by creating i8n-ized sub-sites for the default site.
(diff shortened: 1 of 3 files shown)


therealmaxmoon:
Is having a whole copy of a theme still a solution today, if you support several languages? I have the same problem like [this person](https://github.com/getpelican/pelican/issues/92#issuecomment-1309188) has. The articles are in two languages and can be chosen by a button, which is done really good by i18n_subsites, but if you have a navigation bar in another language and the only way to translate them is to have a copy of a whole theme, only changing some strings, seems a little bit of overhead. 

And because the last post is 6 years old there might be a better solution. 

Thank you for your help :)

PR #167, opened 2014-02-22
Fix sitemap plugin to look for index#.html files (if pagination is on)

Vincent-C:
The current sitemap plugin only adds index.html to the sitemap, but if the user has pagination enabled (which generates index2.html, index3.html, etc.), then these additional index files don't get added to the sitemap.

I've quickly hacked together a solution that works, although I don't think hardcoding "100" in there is the right thing to do. Suggestions?

### DIFF
--- a/sitemap/sitemap.py
+++ b/sitemap/sitemap.py
@@ -206,10 +206,13 @@ class SitemapGenerator(object):
                                                'date',
                                                'url'])
 
-            for standard_page_url in ['index.html',
-                                      'archives.html',
-                                      'tags.html',
-                                      'categories.html']:
+            standard_pages = ['index.html', 'archives.html',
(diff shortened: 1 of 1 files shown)


mknecht:
Hey, what would have to change so that this fix gets included?

justinmayer:
1. a better solution, preferably, than hardcoding "100"
2. clean-up into a single commit

kdeldycke:
This proposed fix is outdated. `sitemap` code has been refactored in commit https://github.com/pelican-plugins/sitemap/commit/c8cf56251bdb6c1e226d9408f620d22b574e2ac0 to make it more sensitive to the site configuration.

I guess this PR can be closed.

justinmayer:
Thanks, Kevin!

PR #189, opened 2014-04-07
liquid_tags.* plugins are not unicode aware

Mihara:
For some of them, like youtube or vimeo, this makes no difference, but for img it does *(you want to be able to write titles in your language)* and for include_code it presents a more complicated problem -- the source code file might not actually be unicode in the first place. *(There are certain cases where this would be the norm.)*

This commit fixes both img and include_code. Tested locally, seems to work.

### DIFF
--- a/liquid_tags/Readme.md
+++ b/liquid_tags/Readme.md
@@ -63,6 +63,12 @@ document:
 
     {% include_code myscript.py [Title text] %}
 
+'chardet' package is required to detect encoding of the file. (Contrary
+to popular belief, not all source code is plain ascii.) You will want to
+install it before using this plugin:
+
+    pip install chardet
+	
 The script must be in the ``code`` subdirectory of your content folder:
 this default location can be changed by specifying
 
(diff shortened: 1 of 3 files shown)


ulikoehler:
Even with the current master `include_code` does not correctly interpret unicode

I worked out a solution for myself (I use only UTF-8):
In `include_code.py`, replace line 119 with:
```python
              + '\n    '.join(code.decode("utf-8").split('\n')) + '\n\n'
```

Maybe someone with more knowledge of the code can work out a solution with chardet?

jcdubacq:
I agree that default to UTF8 plus specifying case by case (for include_code it is necessary) is surely the way to go. `chardet` would be nice but it's not 100% effective, so we would need some kind of manual override anyway.

ulikoehler:
In #550 I have now proposed a full solution with user-selectable codec. This does not require any external libraries, but non-UTF8 users must select the appropriate codec in the `include_code` tag.

ulikoehler:
@Mihara Maybe we can close this pull request now that https://github.com/getpelican/pelican-plugins/pull/550 has been merged ?!? In my opinion #550 fixes all the issues proposed in this pull request in a more generalized way, but please correct me if I'm wrong.

Mihara:
@ulikoehler Well, I certainly don't mind.

PR #210, opened 2014-05-24
[liquid_tags:notebook] Fix hardcoded content path

amic-github:
'content' is not supposed to even be the default for PATH.

Maybe there is a more standard way to do it, but this patch works.


justinmayer:
For now I still think falling back to `content` would be best. Addressed via ce69814.

PR #228, opened 2014-06-27
Allow comparison of TZ and non-TZ timestamps in sitemap plugin

bsidhom:
Date comparisons now allow mixed TZ/non-TZ datetime objects. Previously, if an article, etc., had a date with a time zone but no modification time, a comparison between the date and datetime.min (which has no zone information) caused a crash.

This works well as a short-term fix, but it is preferable to fix the issue within pelican itself. Unzoned date information should be automatically given the time zone from the site configuration.

### DIFF
--- a/sitemap/sitemap.py
+++ b/sitemap/sitemap.py
@@ -52,6 +52,16 @@ def format_date(date):
         tz = "-00:00"
     return date.strftime("%Y-%m-%dT%H:%M:%S") + tz
 
+def date_max(a, b):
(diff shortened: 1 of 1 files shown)


kdeldycke:
Timezone management has been improved and refactored at: https://github.com/pelican-plugins/sitemap/pull/3

This PR is a duplicate and can be safely closed since its bug fix has been moved to the new repository.

justinmayer:
Thanks, Kevin!

PR #267, opened 2014-08-17
FIX: removes the '-None' inserted into output HTML filenames by asciidoc_reader plugin

0xMF:
Removes the '-None' that automatically gets added to HTML files generated by AsciiDoc plugin. Also removes the '-None' from the URL links to those files. 

After this fix, the AsciiDoc plugin behaves the same way as other Pelican readers (like Markdown) do, meaning both the filenames and URL links to those files are named the same way. So if a file is test.asciidoc, the HTML file is now test.html (before this bug fix AsciiDoc files would have been named test-None.html) .


0xMF:
Leave a comment

Issue #312, opened 2014-09-23
Markdown 2.5 breaks _nb_header.html generation in liquid_tags


ghost:
I'm trying to use the liquid_tags.notebook package, and proceeding as instructions indicate on Readme (using Markdown 2.5 package) leads to a "KeyError: 'settings'" in line 12 of liquid_tags.py when I call 

```
make html
```

When I uninstall Markdown 2.5 and install 2.4 instead I don't get this error.  It seems 2.5 just came out on [Sept. 13, 2014](https://pypi.python.org/pypi/Markdown/2.5), so I assume some interface change there is likely the issue.  I suspect it's an easy enough fix, but I'm really not competent enough with Python to figure it out.

In case it helps, pip freeze spits out the following for my virtualenv:
Jinja2==2.7.3
Markdown==2.5
MarkupSafe==0.23
Pygments==1.6
Unidecode==0.04.16
argparse==1.2.1
blinker==1.3
docutils==0.12
feedgenerator==1.7
ipython==2.2.0
pelican==3.4.0
python-dateutil==2.2
pytz==2014.7
six==1.8.0
wsgiref==0.1.2

I'm going to just use the 2.4 for now, but hopefully this doesn't take too long to fix for somebody who knows what they're doing.

PR #321, opened 2014-10-01
Fix liquid_tags for Markdown 2.5

jcdubacq:
This patch closes issue #312 as far as I can see.

Markdown 2.5 changed the config system for markdown extensions. Inspired by
the render_math plugin, I added code to cope with both versions. New settings
used in the markdown part must go in LT_CONFIG and LT_HELP in `mdx_liquid_tags.py`.


naarkhoo:
I still have the same problem but really don't get from hear how to patch it ! I use Markdown==2.5.1 and pelican==3.5.0. along with Python 2.7.6.

PR #322, opened 2014-10-01
Allow utf8 files in include_code liquid tag

jcdubacq:
Nowadays, source files can easily be in utf-8 instead of ascii.


kdeldycke:
The `liquid_tags` plugin has been migrated as a stand-alone plugin at https://github.com/pelican-plugins/liquid-tags . I recommend you to try out that new version as some bugs were fixed and the plugin is actively being worked on over there.

In the mean time I guess we can close this PR.

PR #323, opened 2014-10-01
Fix absolute path instead of relative in liquid_tags/include_code

jcdubacq:
URL of source codes are absolute, instead of being relative
When the site is not at the top of the hierarchy, the URL does not connect to the file


justinmayer:
This PR has unfortunately become out-of-date, with a number of unrelated commits clogging up the works. If anyone wants to propose related follow-up changes, please do so in the context of a new PR.

Issue #338, opened 2014-10-20
asciidoc reader gives encoding error


hjwp:
Tried to do a bit of digging into why, it's not the fault of pelican_open.

Replacing cStringIO with StringIO helps, but then we still get stuck on:

```
FAILED: <stdin>: line 1: unexpected error: 'ascii' codec can't decode byte 0xef in position 0: ordinal not in range(128)
Traceback (most recent call last):
  File "/home/harry/.virtualenvs/www.obeythetestinggoat.com/local/lib/python2.7/site-packages/pelican/generators.py", line 470, in generate_context
    context_sender=self)
  File "/home/harry/.virtualenvs/www.obeythetestinggoat.com/local/lib/python2.7/site-packages/pelican/readers.py", line 446, in read_file
    content, reader_metadata = reader.read(path)
  File "/home/harry/workspace/www.obeythetestinggoat.com/pelican-plugins/asciidoc_reader/asciidoc_reader.py", line 48, in read
    ad.execute(text, content, backend=backend)
  File "/home/harry/workspace/www.obeythetestinggoat.com/pelican-plugins/asciidoc_reader/asciidocapi.py", line 248, in execute
    raise AsciiDocError(self.messages[-1])
AsciiDocError: FAILED: <stdin>: line 1: unexpected error: 'ascii' codec can't decode byte 0xef in position 0: ordinal not in range(128)
```

Which looks to me like it's the actual asciidoc api returning the error.  Ran out of time just now to investigate further...

nickoe:
I just tested the addition of those encode and decode methods, which I can confirm work. I don't know about the meta data thing, I have not tested that. It is bad style to include unrelated fixes for a bug fix. But I think the fix should be comitted, they should still be compatible with python 2 and 3.

justinmayer:
Ostensibly addressed by #453. If the issue persists, please comment here.

PR #339, opened 2014-10-24
New unit test, and hacky fix for, encoding issues (#338)

hjwp:
The fix is hacky, but it's better than the current situation I hope.  Could always merge this, and then look for a less hacky fix...


nickoe:
Could this be pulled in?

ingwinlu:
i advise against a merge because:
* it breaks python3 compability
* sys.setdefaultencoding

PR #348, opened 2014-11-06
Added new plugin: pelican-genealogy

zappala:
This is a small plugin that helps genealogy bloggers. The plugin uses new metadata to track surnames and people mentioned in a blog post. It creates a page for each surname and person listing all the articles where they are mentioned, similar to how tags are handled. The plugin also provides context for themes to get a list of all surnames and people.


justinmayer:
Thanks for the contribution, Daniel. Merged as submodule via https://github.com/getpelican/pelican-plugins/commit/1a0ed58cd1cfe063ef3e549f4937613c0078c170.

PR #358, opened 2014-11-13
Add the pelican_commonmark plugin

niligulmohar:
This provides a reader for markdown files that uses the Python parser
for CommonMark.

### DIFF
--- a/pelican_commonmark/README.md
+++ b/pelican_commonmark/README.md
@@ -1,0 +1,18 @@
+# CommonMark for Pelican
+
+This reader plugin replaces the markdown reader with one that uses the
+[Python parser for CommonMark][1].
+
+It is useful if you want to use the [CommonMark][2] way of parsing markdown
+inside non-trivial nested HTML tags. It is not useful if you want to
+use the extensions available to the python markdown parser.
+
+## Requirements
+
+The plugin uses the CommonMark python package. It can be installed
+with the following command:
+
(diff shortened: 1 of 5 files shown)


kylef:
Looking at this implementation, I don't think this would actually work. You don't seem to register the newly created reader. Take a look at the documentation on creating a reader: http://docs.getpelican.com/en/latest/plugins.html#how-to-create-a-new-reader

It would probably also make sense to un-register the markdown reader too.

niligulmohar:
I found it was only necessary to have a `file_extensions` property, and to subclass `BaseReader`. `Readers.__init__` then handles registering the reader.

But perhaps I shouldn't rely on this behavior?

niligulmohar:
Suggested improvement:  69891dd

Now all extensions for the `MarkdownReader` are changed to the `CommonMarkReader` through the `readers_init` signal.

avaris:
It's true that if you subclass `BaseReader`, you won't need to explicitly add the reader. But this relies on the order of `Reader` object instantiation and the import of your plugin. If import doesn't come first, it won't work. Currently, pelican will import plugins first, but it's not guaranteed that it'll be like this in the future.

justinmayer:
Given the inactivity, I am closing this pull request for now.

@niligulmohar: If you would like to continue working on your CommonMark plugin, please post a comment here, and I will re-open.

PR #377, opened 2014-12-19
Added 'current_lang' to context

joaobarbosa:
Adding the current language to the context was useful for language-related conditions, like display some content only when a specific language is being displayed.


joaobarbosa:
@smartass101 I came to this in the following situation. I was working on a project with different languages, and I had to display some content just in some of them. I got stuck when I tried to get the current language to check if I would have to display the specific content, and the `DEFAULT_LANG` was always the same. Then, I created this `current_lang`, so I was able to check the language being displayed, apart of the default one. (And, that's true the `DEFAULT_LANG` gets duplicated when displaying the default website language).

joaobarbosa:
@smartass101 @justinmayer just did some tests, and now the default lang is working properly, I guess. I was a misunderstood. Thanks :)

PR #378, opened 2014-12-19
Fix to make filetime_from_git work with GitPython 0.3.2.1

skempken:
When using the _filetime_from_git_ plugin with GitPython 0.3.2.1 installed, I encountered the following error message: 
```
 'Repo' object has no attribute 'commits'
```

This is just a quick fix to make it work with the most recent version of GitPython. When I run pelican now, it works as described, but warns about

```
Exception ImportError: ImportError('No module named sys',) in <bound method WindowCursor.__del__ of <smmap.mman.WindowCursor object at 0x10a2de368>> ignored
```

However, this is something that is either related to my environment or to smmap as a dependency. Nevertheless, I'd be happy to get any hints on how to fix it. :-)

### DIFF
--- a/filetime_from_git/filetime_from_git.py
+++ b/filetime_from_git/filetime_from_git.py
@@ -45,7 +45,7 @@ def filetime_from_git(content):
         content.date = datetime_from_timestamp(os.stat(path).st_ctime, content)
     else:
         # file is managed by git
-        commits = repo.commits(path=path)
+        commits = list(repo.iter_commits(path=path))
         if len(commits) == 0:
             # never commited, but staged
             content.date = datetime_from_timestamp(os.stat(path).st_ctime, content)


cscutcher:
Can I ask is there a reason that this hasn't been merged yet? I just hit the same issue.

justinmayer:
@cscutcher: Did this fix resolve your issue? Without any warnings or side effects?

cscutcher:
Since you put me on the spot I did a bit more tinkering and I don't think this pull request can entirely solve the problem. While this will fix things for gitpython 1.0.0 it'll break it for older versions. 

We can't just insist on using the newer (or older) version as it seems versions of gitpython have a dependency on the version of the binary git on a system.

I have proposed pull request #473 which I think will solve the issue in a more backwards compatible way.

justinmayer:
PR 473 merged in favor of this one.

@skempken: If you think there is more to do here, please feel free to submit a follow-up pull request. Thanks!

PR #382, opened 2014-12-25
Replace pelican-md-metayaml with an updated version

pR0Ps:
Replaces the pelican-md-metayaml plugin with an updated fork of it.

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -126,6 +126,8 @@ Read More link            Inserts an inline "read more" or "continue" link into 
 
 Related posts             Adds the ``related_posts`` variable to the article's context
 
+Markdown-metaYAML         Pelican reader to enable YAML-style metadata in markdown articles
+
 Math Render               Gives pelican the ability to render mathematics
 
 Panorama                  Creates charts from posts metadata


kernc:
How is your updated version better? What is updated? The commit message seems pretty insufficient.

Python-Markdown supports YAML metadata from version 2.6+, but Pelican is as of yet incompatible with it.

pR0Ps:
I added more detailed information to the commit.

The reason I started looking into this plugin is that it didn't work (for me anyway). In fixing it, I noticed that it was inheriting the `BaseReader` class and re-implementing a bunch of logic already defined in the `MarkdownReader` class. It was also doing some crazy pathing hacks to load the markdown yaml plugin (which was an outdated version because the source was just pasted in at some point).

I basically rewrote the plugin, taking out the pathing hacks, reusing functionality already built into the `MarkdownReader` class, and changing the library it uses to be a submodule pointing at the current version of the project.

pR0Ps:
@justinmayer: Could you take a look at this and let me know if there are any issues with it?

justinmayer:
Since Python-Markdown 2.6 will support YAML metadata, it probably makes the most sense to build on top of that.

@waylan: Do you think it might be possible to release 2.6 sooner rather than later?

kernc:
I'd be happy to hack the missing bits for seamless support in Pelican. From top of the head, I think the only two issues are:
* `METADATA_PROCESSORS['date']` needs to support date already being a `datetime` object, and
* [MarkdownReader](https://github.com/getpelican/pelican/blob/88ec7026ea493789f91c8bebbcfb283b10032061/pelican/readers.py#L202) should relax it's `meta` requirement to allow `meta(yaml=True)`

pR0Ps:
I've submitted [a pull request to the original project](https://github.com/joachimneu/pelican-md-yaml/pull/5) with these changes in it, hopefully it's accepted. I didn't submit one initially because in my experience, completely rewriting someones code and asking them to replace it doesn't usually go over well.

If/when that pull request is accepted, I'll update this pull request to update to it instead. If Markdown is updated to 2.6 and works properly with Pelican before then, this issue can be closed.

waylan:
I expect that Python-Markdown 2.6 will be released relatively soon. Possibly within a week - time permitting.

pR0Ps:
The pull request to the original project was accepted so I've updated this pull request to just update the plugin instead of replace it with a different project.

justinmayer:
Thanks, Waylan. Looking forward to the release.

And thanks for the plugin rewrite, Carey!

kernc:
@justinmayer Looking forward to having Pelican 3.6 deferred by then or at least until the support is in there (working on it).

waylan:
FYI, I just released [Python-Markdown 2.6][1]. See the [release notes][2].

[1]: https://pypi.python.org/pypi/Markdown/2.6
[2]: https://pythonhosted.org/Markdown/release-2.6.html

justinmayer:
@waylan: Many thanks for releasing 2.6 within 24 hours of my request. You rock!  :+1:

@kernc: Are you working on support for YAML-based metadata for Markdown in Pelican? That's how I interpreted your comment above, but I didn't want to assume as much. (^_^)  Let me know either way, as I'm trying to plan the schedule for the Pelican 3.6 release. Thanks!

kernc:
@justinmayer, I absolutely am/have been. It just took a little longer than first expected as something more urgent came up. Here's an attempt: https://github.com/getpelican/pelican/pull/1636 Also can look into migrating the 2.6-introduced deprecations if that'd be desirable.

waylan:
FYI, you guys might want to see the discussion going on [here][1] before going any further with using the built-in meta-data extension for YAML support.

[1]: https://github.com/waylan/Python-Markdown/issues/390

Issue #39, opened 2013-06-27
Have a catalog of all the plugins


almet:
Sublime text, for instance, maintains a simple json file hosted somewhere remotely and there is a little command line to list / install the plugins. I suppose we could do this.

It means that each plugin should have a setup.py and proper packaging, but that would also enable people to not depend on this pelican-plugins repository.

Thoughts?

justinmayer:
Thanks, @mitchtbaum. I made a formal request via that channel: https://bitbucket.org/pypa/pypi/issue/238/request-for-new-pypi-trove-classifier-for

If Richard has not responded in a day or two, feel free to post a comment there in support of this initiative.  (^_^)

mitchtbaum:
How about using [peru](https://github.com/buildinspace/peru)?  AFAICT, it would satisfy our needs for both plugins and themes, provide a robust and pluggable solution for future extensibility, and it's written in Python.  On top of that, contributors would be able to continue with their existing workflow, safe from any additional efforts with pypi's packaging requirements.  It seems like a win win win.

mitchtbaum:
Moved previous comment to https://github.com/getpelican/pelican/issues/1655

In terms of a catalog, perhaps it would work well for us to have a flat-file database of pointers to each plugin and theme, and even to a DOAP file within them.  That way, a Pelican site, possibly running on getpelican.com, on build, could HTTP request that file and XPath query its description, ie https://github.com/edumbill/doap/blob/master/examples/doap-doap.rdf#L10 .  We would have nicely manageable version control and collaborative access for updates when people want to add or change references to this database.

mitchtbaum:
A little more info on using DOAP files:
* [Apache Software Foundation on DOAP files](http://projects.apache.org/doap.html)
* [GNOME GIT FAQ: How do I add a description to the git web view? What is this "blah.doap"?](https://wiki.gnome.org/Git/FAQ#How_do_I_add_a_description_to_the_git_web_view.3F__What_is_this_.22blah.doap.22.3F)

mitchtbaum:
Upon further analysis, I recommend against using DOAP files.  DOAP is dopey, not dope. :-p

Seriously though, I have 2 reasons: 1) markdown is way better for file-based, structured data entries (ie Pelican's metadata field support), and 2) it overloads a bunch of important pieces of info into a single file and forces stale redundancy with data sources such as README's full description and VCS's latest tag.

[...]

PR #388, opened 2015-01-02
Assets: Allow to explicitly set debug mode

robclancy:
Allow to explicitly set debug mode instead of relying on the logger. Running `--debug` in pelican only changes the logging so it can be used in production however with this using the logger to change how assets work I can no longer use it in production (which is needed for other reasons). So this change is to allow more control.

I haven't tested or anything, about to make the change myself and use it. However adding code here should easily show you what  I mean if you want to implement it a different way.

### DIFF
--- a/assets/assets.py
+++ b/assets/assets.py
@@ -51,7 +51,9 @@ def create_assets_env(generator):
         for name, args, kwargs in generator.settings['ASSET_BUNDLES']:
             generator.env.assets_environment.register(name, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


robclancy:
This single repository for all these plugins is a terrible idea. I'll just fork this instead like I have had to do with others because of this bizarre setup.

justinmayer:
We have already acknowledged that this could (and will) be improved, as noted in #39.

robclancy:
Thank you.

PR #391, opened 2015-01-05
org_reader plugin. Publishes Emacs Org-mode files.

ardekantur:
org_reader plugin, for reading Emacs Org-mode files for publishing as articles in Pelican.


justinmayer:
Thanks for the contribution, Matthew!

PR #392, opened 2015-01-09
Event listing, alpha stage.

FedericoCeratto:
A plugin to generate an events list and a vCalendar file.


justinmayer:
Thanks, Federico!

PR #393, opened 2015-01-10
Updated html snippet of youtube

Azd325:
Updated the youtube url to get the embedded player to work again

### DIFF
--- a/liquid_tags/youtube.py
+++ b/liquid_tags/youtube.py
@@ -14,7 +14,7 @@ Example
 
 Output
 ------
-<iframe width="640" height="480" src="http://www.youtube.com/embed/dQw4w9WgXcQ" frameborder="0" webkitAllowFullScreen mozallowfullscreen allowFullScreen></iframe>
+<iframe width="640" height="480" src="https://www.youtube.com/embed/dQw4w9WgXcQ" frameborder="0" webkitAllowFullScreen mozallowfullscreen allowFullScreen></iframe>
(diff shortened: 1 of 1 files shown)


claman:
I tried making these changes, but they don't seem to work.

I'm using Safari 8.0.2 on OS X 10.10.1

Azd325:
@claman Do you mean the relative urls?

claman:
@Azd325 sorry, probably should have commented on the code itself; I mean the change from `src='http://www.youtube.com/v/{youtube_id}'` to `src='//www.youtube.com/embed/{youtube_id}'`

Azd325:
@claman I don't have a safari. Can you check which embed code is shown on the youtube page if you click share. In chrome and firefox is the default with "//" only

claman:
@Azd325 it's with "//" only

justinmayer:
Thanks for the submission. As mentioned on Paul Irish's [treatise on protocol-relative URLs](http://www.paulirish.com/2010/the-protocol-relative-url/), this technique is now deemed an anti-pattern. I can't think of any reason not to use HTTPS for all YouTube links.

After making those changes, please squash your commits. Thanks!

Azd325:
@justinmayer Changes are made. 

Maybe we should close than #393. If we don't want to use this anti-pattern in the liquid-tags.

justinmayer:
Thanks, Tim!

PR #395, opened 2015-01-10
Adjust rules for what gets compressed by `gzip_cache`.

zackw:
* Don't compress WOFF files.  This font format has internal compression;
   gzip can often shave a few more bytes off but it's not worth the overhead.

 * For any type of file, only save the compressed version if compression
   actually made it smaller.


justinmayer:
Excellent work. Thanks, Zack.

PR #397, opened 2015-01-13
adds instruction message

digitalvapor:
adds a simple instruction message. useful for neophytes who might confuse w3c_validate vs wc3_validate from the `wc3_validate.py` file vs the plugin name.

PR #400, opened 2015-01-13
adds instruction to w3c_validate

digitalvapor:
adds a simple instruction message. useful for neophytes who might confuse w3c_validate vs wc3_validate from the `wc3_validate.py` file vs the plugin name.

re-do of pull request https://github.com/getpelican/pelican-plugins/pull/397 for neatness.


digitalvapor:
thanks!

PR #404, opened 2015-01-15
add toc plugin

ingwinlu:
Differences between pelican-toc and pelican-extract-toc

`extract-toc` uses a markdown extension to generate a toc and then extract it via beautifulsoup. This extension generates the toc itself, removing the need to write `[ToC]` in your articles.


ingwinlu:
going over my old open PR's, rebased of current master

justinmayer:
Thanks, @ingwinlu!

PR #405, opened 2015-01-16
Panorama v0.1

romainx:
Hello,

I've created a plugin, called Panorama, to generate statistics from blog posts (number of posts per year, per categories and so on) and display them as beautiful charts.

- Statistics are made with the [Pandas Data Analysis Library](http://pandas.pydata.org)
- Charts are built with the [Python Wrapper for NVD3](https://github.com/areski/python-nvd3)
- A live example can be seen [here](http://aubonroman.com/stats.html).

I've built this plugin for my needs -- and for fun. It is not perfect but it may be useful to someone else -- or someone can contribute to improve it. Let me know.

Thanks,
Romain

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -58,3 +58,6 @@
 [submodule "pelican-md-yaml"]
 	path = pelican-md-yaml
 	url = https://github.com/joachimneu/pelican-md-yaml.git
+[submodule "panorama"]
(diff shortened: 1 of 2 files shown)


justinmayer:
Thanks for the contribution, Romain!

PR #406, opened 2015-01-27
[interlinks] support img src attr too

andyli:
Such that ppl can do:
```
![Site Logo](this>logo.png)
```

It could be a temp workaround for https://github.com/getpelican/pelican/issues/812


justinmayer:
Many thanks for the enhancement, Andy. Very sorry it's taken so long to merge!

PR #407, opened 2015-01-29
Reading external files using include_code tag from liquid_tags crashes with following letters: ç, á, ó..

thiagogomesverissimo:
when external code has letters like ç, á, ó etc (brazilian case) include_code tag from liquid_tags crashes. Set to be always utf-8. I do not know if it is the best solution, however it worked well for me.

### DIFF
--- a/liquid_tags/Readme.md
+++ b/liquid_tags/Readme.md
@@ -80,7 +80,7 @@ only allowed if a title is also given.
     {% include_code /path/to/code.py lines:1-10 :hidefilename: Test Example %}
 
 This example will show the first 10 lines of the file while hiding the actual
-filename.
+filename. UTF-8 encoding is assumed for code files.
 
 The script must be in the ``code`` subdirectory of your content folder:
 this default location can be changed by specifying
--- a/liquid_tags/include_code.py
+++ b/liquid_tags/include_code.py
@@ -33,10 +33,6 @@ import re
 import os
 from .mdx_liquid_tags import LiquidTags
 
-import sys
-reload(sys) 
-sys.setdefaultencoding('utf8')
-
 SYNTAX = "{% include_code /path/to/code.py [lang:python] [lines:X-Y] [:hidefilename:] [title] %}"
 FORMAT = re.compile(r"""
 ^(?:\s+)?                          # Allow whitespace at beginning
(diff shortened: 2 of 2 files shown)


avaris:
Please, no. Python removes `sys.setdefaultencoding` at startup because you're not supposed to change it during execution (That's why you had to do `reload(sys)` just to use it). Relying on implicit encoding/decoding is also not ideal. Besides, this won't work on Python 3 (`sys.setdefaultencoding` no longer exists and `reload` is not a builtin).

Below is the diff for a proper solution. Since you already opened a PR, I didn't want to create a new one. Can you update yours? Also, it might make sense to update the `Readme.md` file to state that `utf-8` encoding is assumed for code files.
```diff
index cdb6377..73e79e2 100644
--- a/liquid_tags/include_code.py
+++ b/liquid_tags/include_code.py
@@ -29,6 +29,9 @@ in the STATIC_PATHS setting, e.g.:

 [1] https://github.com/imathis/octopress/blob/master/plugins/include_code.rb
 """
+from __future__ import unicode_literals
+
+from io import open
 import re
 import os
 from .mdx_liquid_tags import LiquidTags
@@ -77,7 +80,7 @@ def include_code(preprocessor, tag, markup):
     if not os.path.exists(code_path):
         raise ValueError("File {0} could not be found".format(code_path))

-    with open(code_path) as fh:
+    with open(code_path, encoding='utf-8') as fh:
         if lines:
             code = fh.readlines()[first_line - 1: last_line]
             code[-1] = code[-1].rstrip()
```

thiagogomesverissimo:
Sorry, I do not know python very well, I only use pelican, however, thank you for your advices, I learned a lot. I committed the changes.

avaris:
You forgot to add the `import` at the top. They are needed.

    from __future__ import unicode_literals

    from io import open

And after that, if you can [squash](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits) your commits to get a single one, it'd be perfect.

thiagogomesverissimo:
I tested without this imports and it worked, that is why I committed without this lines before. Now I did a new commit with the import.

thiagogomesverissimo:
I tested again and it is necessary include the following lines:
from __future__ import unicode_literals
from io import open

Thank you so much!

PR #412, opened 2015-02-01
Added option to save the file tree in the generation of thumbnailes.

Iv:
May be not final version, but works for me and rather usefull.


justinmayer:
Thanks for the contribution, Ivan!

PR #414, opened 2015-02-05
Add draftfuture plugin.

ionelmc:


justinmayer:
Isn't this functionality already present in Pelican core? Try:

    WITH_FUTURE_DATES = False

If this plugin does something different, then by all means please let me know, of course.  (^_^)

ionelmc:
Ooops. I've tried to use that now but I get `can't compare offset-naive and offset-aware datetimes`. Perhaps I should fix that instead ...

This plugin does set the list of drafts in the context (seems it's not generated by default).

PR #418, opened 2015-02-07
Headerid

ionelmc:


justinmayer:
Hi Ionel. Are you targeting Markdown? If so, I believe Python-Markdown supports this already via its [TOC extension](https://pythonhosted.org/Markdown/extensions/toc.html).

I'm less familiar with reST, but I believe it also supports header ID generation.

ionelmc:
Hey @justinmayer, this is for reStructuredText. And no, it doesn't have any builtin headerid thing, unlike some of the markdown libs. You can see it in action [here](http://blog.ionelmc.ro).

justinmayer:
Oh, I see now that you did indeed mention that in the ReadMe commit. Thanks!

justinmayer:
Clarified things a bit in the README via f7174a2.

PR #421, opened 2015-02-11
render_math responsive mode

barrysteyn:
The mathjax plugin does not behave well for responsive sites. This is a difficult problem to solve fully, but it can be alleviated by setting automatically setting line-breaks, aligning displayed math to the left and ensuring no indent.

This change allows a user to specify if repsonsive should be attempted (defaults to false for backwards compat) . It also allows the user to set the responsive breakpoint


justinmayer:
Thanks, Barry!

PR #424, opened 2015-02-28
Allow render-math plugin to work for reST if Markdown is not installed

fbeutel:
When markdown is not installed on the system, the render-math plugin doesn't work because it tries to import markdown.

I added exception handling so that users who don't use markdown (yet) can still use this plugin.

PS: I hope it is ok to create the pull request here instead of @barrysteyn 's personal repository...

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -31,7 +31,13 @@ import os
 import sys
 
 from pelican import signals
-from . pelican_mathjax_markdown_extension import PelicanMathJaxExtension
+
+try:
+    from . pelican_mathjax_markdown_extension import PelicanMathJaxExtension
+except ImportError as e:
+    PelicanMathJaxExtension = None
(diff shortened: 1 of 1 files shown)


barrysteyn:
@fbeutel I think this is a great addition, but can you please update my "unofficial" on my personal repo. The reason is because I use that repo to try test as many scenarios as possible. I have plenty of "math and physics customers" for this plugin, and as soon as something breaks, they seem to email me personally (Note: never give your email away!). So if you don't mind updating my personal repo, I will make sure that you get the credit if this is used (and it looks like it should be used)

barrysteyn:
Just reproduced things on my local repo, and everything is good. :+1:

justinmayer:
Thanks to @fbeutel and @barrysteyn for the enhancement and review, respectively!

PR #428, opened 2015-03-04
Allow using the render_math plugin without the markdown module

bugnano:
If the markdown module is not installed, the render_math plugin fails with an ImportError.
This patch allows using the render_math plugin on reStructuredText files, without having to install also the markdown module.

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -31,7 +31,10 @@ import os
 import sys
 
 from pelican import signals
-from . pelican_mathjax_markdown_extension import PelicanMathJaxExtension
+try:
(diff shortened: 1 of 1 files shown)


justinmayer:
Hi Franco. It seems your pull request addresses similar goals as those in #424, which was recently merged. Would you please take a look? If you think there is more to do and would like to build on top of those changes, by all means please do so by rebasing on current master and updating this pull request with a single squashed commit. Thanks!

bugnano:
Hi Justin. Sorry, I didn't look whether there was another pull request addressing the same issue before creating this one.
I think that #424 addresses the issue better than my pull request, so I'm ok with merging that one.

PR #434, opened 2015-03-10
Fix bug in Thumbnailer When IMG_PATH has trailing slash

eupharis:
When ```pelican.settings.IMG_PATH``` has a trailing slash, Thumbnailer chops off the first letter of the thumbnail name. This PR contains a failing test demonstrating the issue:

```
AssertionError: 'sample_image_square.jpg' != 'ample_image_square.jpg' 
```

This PR then fixes the bug.


justinmayer:
Thanks for the fix, Dan!

PR #438, opened 2015-03-12
Only apply workaround for issue 1547 if core exhibits the bug.

zackw:
getpelican/pelican#1547 has been fixed, so don't monkey-patch the Content classes if it isn't necessary.

The workaround can be removed entirely once everyone who cares about this plugin is on a released version with the bugfix.

I was going to write a test suite along with this change but I failed to make unittests and Pelican play nice together, and I wasn't able to find a working plugin testsuite to crib from!  Anyhow, I've manually verified that this still works on the website that I use it for, with a version of Pelican that does *not* have the bug fixed.

### DIFF
--- a/category_meta/category_meta.py
+++ b/category_meta/category_meta.py
@@ -1,4 +1,4 @@
-'''Copyright 2014 Zack Weinberg
+'''Copyright 2014, 2015 Zack Weinberg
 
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks, Zack!

PR #448, opened 2015-03-27
clone submodules too when installing

benjaminabel:


kylef:
This PR also has a change you didn't mention, which is going from https checkout to SSH checkout which then requires the user to have a GitHub account and an SSH key attached.

Can you revert the second change so this can become merged?

benjaminabel:
@kylef Sorry I didn't see that change, you're right, it's better to stay in https.

kylef:
Thanks for updating @benjaminabel.

benjaminabel:
You're welcome

PR #450, opened 2015-03-30
Overhaul Summary plugin for efficiency, Py3k compatibility, less convoluted control flow

zackw:
I implemented all these changes while debugging a problem with my blog ... which turned out to be `read_more_link`'s fault, but it wound up being easier to add a feature to `summary` enabling me not to *need* `read_more_link`, than to fix `read_more_link`.  And I think these changes are worthwhile in their own right.

 * New feature: `instance.has_summary` set to True if an explicitly-defined
   summary exists, False otherwise.  Templates can use this to inject a
   "read more" link for instance (the `read_more_link` plugin seems to be
   broken under Py3k).

 * Tested under Python 3.4; import unicode_literals for consistent behavior
   on both Py2 and Py3.

 * `instance._content` is modified directly instead of using the undocumented
   `instance._get_content` hook.
 * Control flow within `content_object_init` hook is streamlined.
 * Settings are looked up only once per article, which helps both
   efficiency and readability.

### DIFF
--- a/summary/Readme.rst
+++ b/summary/Readme.rst
@@ -26,6 +26,11 @@ beginning marker was found, it starts at the top of the body. It is possible
 to leave out the end marker instead, in which case the summary will start at the
 beginning marker and continue to the end of the body.
 
+The plugin also sets a ``has_summary`` attribute on every article. It is True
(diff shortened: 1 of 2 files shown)


justinmayer:
Thanks, Zack!

PR #452, opened 2015-04-01
Supports sub-parts, which are articles nested below other articles.

pxquim:
Use sub-parts to break a very long article in parts, without polluting
the timeline with lots of small articles. Sub-parts are removed from
timelines and categories, but remain in tag and author pages.


pxquim:
Comments welcome.

justinmayer:
Hi @pxquim. How is this different from the existing [Series plugin](https://github.com/getpelican/pelican-plugins/tree/master/series)?

pxquim:
Sub-parts solves a different problem than Series. In fact, I use both plugins at pxquim.pt.

- Series connects several articles, where each article stands on its own. I use Series in pxquim.pt for events with several shows. While each show stands on its own (it is a different artist or band), I use the Series to state that they all happened in the same place, or roughly at the same time.

- Sub-parts breaks a large article into parts. The first motivation was to break a karate competition, where I easily publish 1000 photos, into several articles, each with its own (smaller) gallery. However, each gallery does not really stand on its own, because I do not have enough information to publish next to each gallery. Also, I did not want an avalanche of sub-standard articles "polluting" the main timelines. With Sub-parts, only the "main" article appears in the timelines, and you navigate to sub-parts from the main article.

See also getpelican/pelican#1617 from last February.

justinmayer:
I see that you already explained this several times, including in the previous issue as well as the plugin's ReadMe. And now here, at my prompting. Please accept my apologies for making you repeat yourself unnecessarily!

Thank you for your work on this contribution.  (^_^)

pxquim:
No problem, @justinmayer, and thanks for the work that you are doing.

Actually, I forgot to mention two examples of using the Sub-parts plugin (in English). I'll add them here as examples for others; I guess they would look out-of-place at the README of the plugin.

http://pxquim.com/blog/2015-02-26-cenjor.html
:   The main article is my experience at the event, with subparts detailing presentations and the event itself. The presentations might be separated "main" articles (using the Series plugin). However, the explanation of the event would look out-of-place on its own, and the presentation by the secretary of state was very shallow (he just talked for a few minutes). So, I published everything as sub-articles.

http://pxquim.com/blog/2014-05-23-apcomtec-aveiro.html
:   The main article describes the day, while the sub-articles explain one presentation and one workshop. These might have used the Series plugin, but the main article provides context for the readers.

PS: Perhaps one day Github will support description lists? What an amazing omission!

PR #454, opened 2015-04-02
Add photos and galleries to articles, resize as needed.

pxquim:
Use Photos to add a photo or a gallery of photos to an article, or to include photos in the body text. Photos are kept separately, as an organized library of high resolution photos, and resized as needed.


justinmayer:
Thanks for the plugin contribution, @pxquim!

three3q:
was looking for that for long.

PR #455, opened 2015-04-04
Add plugin for customized inline HTML for Markdown

barrysteyn:
This plugin allows a user to easily create a customized inline html expression for markdown. Markdown has things like `*` as inline expression, this plugin will allow a user to create whatever new inline expression they want.

I have been using this plugin privately for myself for quite a while, and I realized how useful I find it. So I decided to give it to everyone else. For example, I use it to highlight text in markdown

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -100,6 +100,8 @@ Liquid-style tags         Allows liquid-style tags to be inserted into markdown 
 
 Multi parts posts         Allows you to write multi-part posts
 
+Markdown Inline Extend    Enables you to add customize inline patterns to your markdown
+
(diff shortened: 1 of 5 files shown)


justinmayer:
Thanks for the contribution, Barry. I made some capitalization and phrasing suggestions via inline comments.

Also, would you please add this plugin to the list in the [plugin repo Readme](https://github.com/getpelican/pelican-plugins/blob/master/Readme.rst)?

barrysteyn:
Hi Justin

Done! I have corrected the English, and I added a description to the plugin repo Readme

justinmayer:
Awesome. Thanks, Barry!

PR #459, opened 2015-04-10
Better Images: Prevent errors when RESPONIVE_IMAGES is not set.

adworacz:
This fixes the case where RESPONSIVE_IMAGES is not set and you get
errors like this:

ERROR: Could not process ./my-blog-post.md
  | u'RESPONSIVE_IMAGES'


justinmayer:
Thanks, Austin!

PR #465, opened 2015-04-16
New Plugin: Authorpages

dcrosta:
Requesting feedback on the approach. This basically works right now, but I'm not familiar enough with Pelican to know if I've implemented everything properly, or if I've missed something. All feedback appreciated.

### DIFF
--- a/authorpages/authorpages.py
+++ b/authorpages/authorpages.py
@@ -1,0 +1,39 @@
+import os
+import os.path
+
+from pelican import signals
+
+
+def yield_files(root):
+    root = os.path.realpath(os.path.abspath(root))
+    for dirpath, dirnames, filenames in os.walk(root):
+        for dirname in list(dirnames):
+            try:
+                if dirname.startswith("."):
+                    dirnames.remove(dirname)
+            except IndexError:
+                # duplicate already removed?
+                pass
+        for filename in filenames:
+            if filename.startswith("."):
+                continue
+            yield os.path.join(dirpath, filename)
+
+def test(article_generator):
+    settings = article_generator.settings
+    readers = article_generator.readers
+    path = settings.get("AUTHOR_PAGE_PATH", "authors")
+
+    author_pages = {}
+    for filename in yield_files(path):
+        base_path, filename = os.path.split(filename)
+        page = readers.read_file(base_path, filename)
+        slug, _ = os.path.splitext(filename)
+        author_pages[slug] = page
+
+    for author, _ in article_generator.authors:
+        print "set author.page for %s to %r" % (author.slug, author_pages.get(author.slug, ""))
+        author.page = author_pages.get(author.slug, "")
+
+def register():
+    signals.article_generator_finalized.connect(test)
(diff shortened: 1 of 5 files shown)


authorpages/authorpages.py
```diff
@@ -0,0 +1,39 @@
...
+from pelican import signals
+
+
+def yield_files(root):
+    root = os.path.realpath(os.path.abspath(root))
+    for dirpath, dirnames, filenames in os.walk(root):
+        for dirname in list(dirnames):
+            try:
+                if dirname.startswith("."):
+                    dirnames.remove(dirname)
```
kylef: Not sure what the purpose of this is, it seems `dirnames` is never used?

dcrosta: If you mutate dirnames (at least by removing elements, possibly also by rearranging, etc) then os.walk respects those changes as it continues to iterate the file system. Here in just trying to skip dot-dirs (if any) for efficiency.

kylef: Interesting, I didn't know this. :+1: thanks.

kylef: Interesting, I didn't know this. Nice :+1:

justinmayer:
Thanks for the contribution, Dan.

pxquim:
Very interesting... can we have a similar solution for blog categories? ... and perhaps even blog tags as well?

dcrosta:
@pxquim I don't see why not.

I kind of wish this were a Pelican core feature. Do the core devs have any interest in that? I can probably find some time to try to patch it in there (with support for categories, tags, and authors).

avaris:
I have a plugin in development to handle a similar function for authors, categories and tags for a while now. Unfortunately, pesky real-life issues tend to get in the way. 

Putting it into the core is also a possibility, although I'm sort of against it because I'm in favor of a minimalistic core and a plugin is enough for this task.

dcrosta:
@avaris I just put up #482 which extends the idea here to categories and tags -- you might want to take a look.

avaris:
@drcosta I'm a bit late, it's merged but took a look anyway. 

It's similar to what I'm doing but there are some stuff you might want to watch. 

Authors, categories and tags are not actually singletons. Pelican creates a new object for these types for each article even if there was one with same `name`. `generator.categories` for example would contain only one of them and you're setting `page` for that one. This is mostly OK, but say one wanted to put that stuff in `article.html` template and tried to access `article.category.page`, it might result in an error.

This is not an issue because of the way you set up, but might want to reconsider. I set my paths relative to the `content` so I had to extend `ARTICLE_EXCLUDES` and `PAGE_EXCLUDES` with the paths in order to not create real content from them. Putting these files out of `content` doesn't feel right to me.

Again, this might not be an issue if it's your intention. You're using the `Page` object from core and it'll fire `content_object_init` signal which might trigger some other plugin that's expecting a regular content object. Just something to be careful.

dcrosta:
@avaris thanks, I hadn't considered this:

> You're using the Page object from core and it'll fire content_object_init signal which might trigger some other plugin that's expecting a regular content object. Just something to be careful.

Any suggestions? I wanted to re-use as much code from core as I could, but I don't know core very well yet (I've only browsed the bits that seemed relevant, haven't read it all carefully).

PR #467, opened 2015-04-18
render_math: Python 3, auto insert, and mathjax fonts

barrysteyn:
This plugin is not compatible with Python 3. This PR will fix this problem. It also adds two new features
 
 1. mathjax_fonts - allows the forcing of fonts on the math
 2. auto_insert - allows template designers control over whether the mathjax script is automatically inserted


justinmayer:
Thanks, Barry!

PR #468, opened 2015-04-19
Extend textbundle plugin documentation

DirkR:
I added some more documentation to the textbundle plugin's README and provide a shell function to create new articles in textbundle format.


justinmayer:
Thanks, Dirk!

PR #470, opened 2015-04-20
break out tag_cloud from core into plugin

ingwinlu:
* remove sort_tags plugin
* add tag_cloud plugin with added sorting functions


justinmayer:
Thanks, @ingwinlu!

Issue #384, opened 2014-12-27
filetime_from_git fails with 'Repo' object has no attribute 'commits'


a3nm:
Hi,

When trying to setup filetime_for_git, generation of all files fails with the message "'Repo' object has no attribute 'commits'". I guess the incriminated line is in `filetime_from_git.py`:

    commits = repo.commits(path=path)

Indeed, using the gitpython that I have obtained with `sudo pip install gitpython` on Debian testing (which seems to be version 0.3.2.1 from `/usr/local/lib/python2.7/dist-packages/git/__init__.py` which seems to be the latest version <https://pypi.python.org/pypi/GitPython/>), the `repo.commits` syntax does not work.

I'm assuming that the plugin is expecting a different version of gitpython. In this case, it should perhaps be updated, or at least the exact version of gitpython to use should be specified.

Thanks!

justinmayer:
Addressed via #473 and therefore now closed.

PR #473, opened 2015-04-23
Refactor filetime_from_git and add GIT_FILETIME_FOLLOW option

cscutcher:
The filetime_from_git plugin relies on a now old version of gitpython.
However its not so easy as updating to the latest version as your
gitpython version must match your distros git binary for everything to
work.

This refactor restructures the plugin to provide some isolation from
this issue by wrapping the gitpython module in a simplified wrapper to
handle this difference.

I have also added the GIT_FILETIME_FOLLOW option which allows the
filetime module to follow files across renames.

As far as I can tell this can supersede pull request #378 and fix issue #384

### DIFF
--- a/filetime_from_git/README.rst
+++ b/filetime_from_git/README.rst
@@ -26,6 +26,18 @@ operations like copy, move will not affect the generated results.
 If some article or page doesn't like to use git time, set a ``gittime: off``
 metadata to disable it.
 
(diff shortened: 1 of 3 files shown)


justinmayer:
Thanks for the extensive refactor, @cscutcher.

@skempken: Can you test this pull request and let us know if you have any comments?

skempken:
Thanks for your work! Unfortunately, I won't have time to test this on the short term.

PR #474, opened 2015-04-23
Render math now works properly with summaries

barrysteyn:
This PR does two things:

 1. Inserts the mathjax script so that math in the summaries will be rendered
 2. Restores math that are cut off in the summaries: Pelican can cut math in half when creating its summary. This will result in math not rendering correctly. This PR will fix this state of affairs: If a summary contains math that is cut off, the math will be restored.


justinmayer:
Thanks, Barry!

PR #478, opened 2015-04-27
Add Permalinks plugin. Refactor filetime_from_git plugin.

cscutcher:
I wanted to add a plugin that would allow auto generation of a sort of permalink based on some unchanging information stored in git, so that even after renames or moves, as long as I share the permalink, I'll never end up with link rot.

This plugin generates an additional page for each article and page in a "/permalinks" directory which will point to the current location of the page. The permalink for each page is a base64 encoded hash including the sha of the commit that introduced an article, and it's original filename. 

This includes another quite large refactor of the filetime_from_git refactor and given that it's considerable changed the scope of that plugin I'd understand if you'd prefer it committed as a new plugin, although I've tried to maintain the behaviour and compatibility of the module. 

I've also been a bit naughty in the permalink modules adding helper methods to the content instance during processing of the content_init signal which people might find distasteful. It could be there's a better way to achieve the same thing.

Note that you don't actually have to use the permalink plugin and the git stuff together. You could choose to embed a UUID as permalink_id in the metadata and still get much of the benefit. 

Apologies if I've done anything naive. Still new to pelican. 
Cheers


justinmayer:
Hi Chris. Please accept my apologies for the obscenely long delay in reviewing your pull request. Would you be so kind as to [rebase on current master and squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits) as appropriate? I'll do my best to ensure a quick review this time.  😉

cscutcher:
Hey Justin! No worries! I'd forgotten myself! 

Is commit rewriting (squashing/rebasing) absolutely necessary? I rather avoid it if possible because it'll mean also having to do a bunch of commit rewriting in my other forks and personal branches. 

Cheers
Chris

justinmayer:
Generally we try to avoid having minor cleanup and merge commits cluttering up the repo commit history. As you can see here, there are quite a few of those. But that is my fault after all, so...  Hmm.

cscutcher:
Alright you win :-p although I hate commit rewriting!

justinmayer:
Certainly appreciate your concession here. If you have the time, please find me on the #pelican channel on IRC and let me know more about how this negatively impacts your workflow.

Thanks for the excellent contribution!  (^_^)

PR #481, opened 2015-05-04
Allow for Windows filepaths

FurmanCenterNYU:
The regex used in categories_meta doesn't work with Windows filepaths.


justinmayer:
Thanks for the contribution. Have you tested this on Linux and/or Mac OS X to ensure everything functions as expected?

bkfunk:
We haven't had a chance to test with other systems, no, sorry!

justinmayer:
Someone will have to volunteer to do so before this can be merged. (I'm traveling at the moment and can't right now.)

PR #482, opened 2015-05-07
Extend and rename AuthorPages to AutoPages

dcrosta:
This builds on (and replaces) the Author Pages plugin in #465, and extends the same technique to work for categories and tags (as well as authors, of course).

I also took a moment to replace `print` with `logging`, and added exception handling in the place that seemed most likely to cause exceptions during rendering.


justinmayer:
Thanks for the enhancement, Dan.

PR #485, opened 2015-05-10
Remove textbundle plugin

DirkR:
When I started implementing the textbundle plugin, Pelican was at version 3.4. Meanwhile version 3.5 was released and pelican had a new method "{attach}..." for attaching static files (as I recap).

While writing some more documentation for this plugn recently, I realized that the plugin is no longer needed. The purpose of the textbundle spec (and of my plugin as well) is to organize Markdown files (blog posts and pages) and their attached assets in dedicated folders.

This plugin can be safely deleted. I tested it in my personal blog (https://github.com/DirkR/niebegeg.net)


DirkR:
Hi, are there any daubts to merge this PR, or can I do anything to make it easier for you to evaluate this PR?
The textbundle plugin really is a no-op plugins since Pelican 3.5.

PR #490, opened 2015-05-13
Add the Libravatar plugin

rlaboiss:
Please add my [Libravatar](http://www.libravatar.org) plugin the the pelican-plugins repository.  This plugin is an improvement on the [gravatar](https://github.com/getpelican/pelican-plugins/tree/master/gravatar) plugin.  It is also a replacement for it, since Libravatar redirects to Gravatar when the user profile is not found in its database. As an improvement on the gravatar plugin, the libravatar plugin accepts configuration variables for setting the default for missing images and the size of the picture. It comes also with a unit testing script.


rlaboiss:
In an afterthought, I think that it will be better to maintain this plugin in a separate repository and define a submodule for the pelican-plugins repo.  I am hereby closing this pull request.

PR #496, opened 2015-05-19
Update for tipue search 5

fernandezcuesta:
As per [official documentation](http://www.tipue.com/search/docs/?d=20) `loc` field has been renamed to `url`.


justinmayer:
Thanks for the enhancement!

fernandezcuesta:
You're welcome.
Actually I realized it doesn't matter if you keep both 'url' and the 'loc' fields for backwards compatibility with TS4.

PR #498, opened 2015-05-20
Sitemap: ability to exclude URLs from a sitemap

alexleventer:
Adds the ability to exclude a URL for being added to a sitemap (very common need).


justinmayer:
Thank you for the contribution, Alex. Do you think the plugin's ReadMe needs to be updated with any relevant information?

alexleventer:
Oops didn't mean to create another pull request but the readme changes can be seen here: https://github.com/getpelican/pelican-plugins/pull/499

PR #500, opened 2015-05-22
libravatar: Encode email address before hashing

rlaboiss:
With Python 3, the libravatar plugin yields the following error:

    Unicode-objects must be encoded before hashing

This is caused by the fact that the email string obtained from the header of the content file is stored as a Unicode object in the metadata list in Python 3.  The string must then be encoded before being passed to the `hashlib.md5` function.  According to the current standard, only ASCII characters should appear in email addresses, so encoding `metadata['email']` into ASCII will make the plugin work with both Python 2 and Python 3.


justinmayer:
Thanks, Rafael!

PR #503, opened 2015-05-26
org_reader: Fix JSON-related error

### LINKED ISSUE #502: org_reader json complaints
Using org_reader I get 

```
ERROR: Could not process test.org
  | the JSON object must be str, not 'bytes'
```

the following patch seems to fix it

```
diff --git a/org_reader/org_reader.py b/org_reader/org_reader.py
index efad620..a847ffe 100644
--- a/org_reader/org_reader.py
+++ b/org_reader/org_reader.py
@@ -68,7 +68,7 @@ class OrgReader(readers.BaseReader):
 
         LOG.debug("OrgReader: running command `{0}`".format(cmd))
 
-        json_result = subprocess.check_output(cmd)
+        json_result = subprocess.check_output(cmd, universal_newlines=True)
         json_output = json.loads(json_result)
 
         slug, e = os.path.splitext(os.path.basename(filename))
```

zakkak:


justinmayer:
Thanks for the fix, @zakkak. In the future, a more descriptive commit message would be helpful.

zakkak:
You are right. I should have included the text from the Issue in the
commit message.

justinmayer:
No worries. Thanks again for the contribution.

PR #510, opened 2015-06-04
Add txt2tags reader plugin

cctags:
Would you mind if I add this txt2tags reader into the 'pelican-plugins' repository? I don't find any existing plugin for this. Please correct me if I am wrong.


justinmayer:
Thanks for the contribution, @cctags.

Issue #508, opened 2015-06-02
add linkedin option to share_post plugin


nicchub:
I'll add linkedin option to share_post plugin.

nicchub:
https://github.com/getpelican/pelican-plugins/pull/511

justinmayer:
Added via above-mentioned pull request.

PR #511, opened 2015-06-12
share_post plugin: add share-to-linkedin support

nicchub:
added share to linkedin support, I'll add some new features like enable/disable links and order settings (all optional with default values) via pelicanconf.py

salu2.


nicchub:
Feature: https://github.com/getpelican/pelican-plugins/issues/508

justinmayer:
Thanks for the enhancement, Nicolas.  (^_^)

PR #512, opened 2015-06-13
Updating rmd_reader to use runtime pelican's settings for Markdown

wilsonfreitas:


justinmayer:
Hi Wilson. Thanks for the contribution. As you can see, there are a bunch of noisy commits in your PR. As noted in our contributing guidelines, it's best to keep commits semantic, which in this case is probably a single commit.

This is where I ask you to squash your commits. However, you've made your changes on the `master` branch instead of a new feature branch, which will make rebasing and squashing your commits harder than it should be.

Please review the contributing guidelines thoroughly and ask on IRC if you have any questions about them. Looking forward to your changes so we can get your PR merged.

avaris:
Also, most of this is pretty much unnecessary. Readers will be initialized with the [settings](https://github.com/getpelican/pelican/blob/master/pelican/readers.py#L107). You don't need to extract settings from `pelican_init`. No need for the "load it last" part as well. Just use `self.settings`.

wilsonfreitas:
@justinmayer ok I will do my best to clean up all that mess. But I don't where to start, do you have any suggestion? 

@avaris Thanks for your comment, I've changed my code and everything worked fine.

justinmayer:
@wilsonfreitas: https://github.com/getpelican/pelican/wiki/Git-Tips

PR #513, opened 2015-06-17
org_reader: Correct date format info in README

dsteuer:
It is necessary to use a date format org understands. Therefore the `<>` around the date. Otherwise it won't be recognized as org-property and the date will always be set to 11 January 2030.


justinmayer:
Thanks, @dsteuer.

PR #515, opened 2015-06-20
Add pelican_javascript plugin

mortada:


avaris:
Can you add your plugin (and its description) to the list in [Readme](https://github.com/getpelican/pelican-plugins/blob/master/Readme.rst)?

mortada:
@avaris Sure I've updated the Readme

justinmayer:
Thanks for the contribution, @mortada.

PR #516, opened 2015-06-20
address #514 by article postprocessing after internal links are resolved

### LINKED ISSUE #514: render_math with process_summaries breaks static file inclusion
Took me a while to notice because it's a non-fatal error for pelican, and my web cache kept giving me non-existent files...

However, I've traced a static media bug to render_math.

If you do NOT have the following setting in your pelicansettings.py

````
MATH_JAX = {
    'process_summary': False
}
````
then your blog will have no media included, i.e. ``{filename}`` and ``{attach}`` references to files are not resolved and the media not included.

This seems to be because the math.process_summary method causes pelican to resolve prematurely, such that pelican.generators.StaticGenerator is never instantiated.
Looks like this process_summary feature should be invoked by a different signal?
In any case, if you turn it process_summary, render_math it works just fine.
At least, it should be disabled by default if it has these flakey side-effects.

running a traceback from when the spurious missing file problem is logged, it turns out that media replacement is bring invoked ultimately by render_math.math_process_summary(); However it shoudl have already happened by the time that render_math wishes to summarise it.

In the mean time, turning off process_summary is a workaround.
...in fact, I don't use summaries myself, so I regard that as a fix; however, others might have different requirements ;-)

### PR DESCRIPTION
**Pull request by howthebodyworks**
Cleaner patch to resolve #514 .

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -30,7 +30,7 @@ See README for more details.
 import os
 import sys
 
-from pelican import signals
+from pelican import signals, generators
 
 try:
     from bs4 import BeautifulSoup
@@ -202,7 +202,7 @@ def process_summary(article):
         article._summary = "%s<script type='text/javascript'>%s</script>" % (summary, process_summary.mathjax_script)
 
 def configure_typogrify(pelicanobj, mathjax_settings):
-    """Instructs Typogrify to ignore math tags - which allows Typogfrify
+    """Instructs Typogrify to ignore math tags - which allows Typogrify
     to play nicely with math related content"""
 
     # If Typogrify is not being used, then just exit
@@ -240,7 +240,8 @@ def process_mathjax_script(mathjax_settings):
     """Load the mathjax script template from file, and render with the settings"""
     
     # Read the mathjax javascript template from file
-    with open (os.path.dirname(os.path.realpath(__file__))+'/mathjax_script_template', 'r') as mathjax_script_template:
+    with open (os.path.dirname(os.path.realpath(__file__)) 
+            + '/mathjax_script_template', 'r') as mathjax_script_template:
         mathjax_template = mathjax_script_template.read()
 
     return mathjax_template.format(**mathjax_settings)
@@ -270,8 +271,10 @@ def mathjax_for_rst(pelicanobj, mathjax_script):
     rst_add_mathjax.mathjax_script = mathjax_script
 
 def pelican_init(pelicanobj):
-    """Loads the mathjax script according to the settings. Instantiate the Python
-    markdown extension, passing in the mathjax script as config parameter
+    """
+    Loads the mathjax script according to the settings.
+    Instantiate the Python markdown extension, passing in the mathjax
+    script as config parameter.
     """
 
     # Process settings, and set global var
@@ -295,29 +298,46 @@ def pelican_init(pelicanobj):
     if mathjax_settings['process_summary']:
         process_summary.mathjax_script = mathjax_script
 
-def rst_add_mathjax(article):
-    """Adds mathjax script for RST"""
+def rst_add_mathjax(content):
+    """Adds mathjax script for reStructuredText"""
 
     # This is only value for .rst files
-    _, ext = os.path.splitext(os.path.basename(article.source_path))
+    _, ext = os.path.splitext(os.path.basename(content.source_path))
     if ext != '.rst':
         return
 
     # If math class is present in text, add the javascript
-    if 'class="math"' in article._content:
-        article._content += "<script type='text/javascript'>%s</script>" % rst_add_mathjax.mathjax_script
+    if 'class="math"' in content._content:
+        content._content += "<script type='text/javascript'>%s</script>" % rst_add_mathjax.mathjax_script
 
-def parse_articles(article_generator):
-    """Adds mathjax script to RST and processes summaries"""
+def process_rst_and_summaries(content_generators):
+    """
+    Ensure mathjax script is applied to RST and summaries are
+    corrected if specified in user settings.
+    
+    Handle content attached to ArticleGenerator and PageGenerator objects,
+    since we don't know how to handle other Generator types.
+    
+    For reStructuredText content, examine both articles and pages.
+    If article or page is reStructuredText and there is math present,
+    append the mathjax script.
 
-    for article in article_generator.articles:
-        
-        rst_add_mathjax(article)
-
-        if process_summary.mathjax_script is not None:
-            process_summary(article)
+    Also process summaries if present (only applies to articles)
+    and user wants summaries processed (via user settings)
+    """
+    
+    for generator in content_generators:
+        if isinstance(generator, generators.ArticlesGenerator):
+            for article in generator.articles:
+                rst_add_mathjax(article)
+                #optionally fix truncated formulae in summaries.
+                if process_summary.mathjax_script is not None:
+                    process_summary(article)
+        elif isinstance(generator, generators.PagesGenerator):
+            for page in generator.pages:
+                rst_add_mathjax(page)
 
 def register():
     """Plugin registration"""
     signals.initialized.connect(pelican_init)
-    signals.article_generator_finalized.connect(parse_articles)
+    signals.all_generators_finalized.connect(process_rst_and_summaries)

### CODE REVIEW

**Comment by justinmayer** [comment]
@barrysteyn: Any comments on this pull request?

**Comment by avaris** [comment]
I'm sort of jumping in but... This will fail if any other plugin adds its own generator.

**Comment by howthebodyworks** [comment]
@avaris Could you supply more information? Are you referring to the unpatched version, or perhaps noting that the patch adds *more* problems than those in #514? What kind of failure?

**Comment by avaris** [comment]
@howthebodyworks,

```python
article_generator, pages_generator, static_generator = content_generators
```

This assumes there are only 3 generators are present, which won't be the case in the situations I described above and you'll get an error like:

```python
ValueError: too many values to unpack
```

justinmayer:
@avaris: Please, please *always* feel free to jump in any time.  (^_^)

howthebodyworks:
Good point @avaris.

I'm note really sure how to handle the problem of "generally" injecting mathjax javascript into any generator's content; so the next patch, 5fe29fc, now just looks for `pages` or `articles` attributes in *all* the generators, and assumes that anything stored in those attributes would like to have mathjax injected.

This works for my site at least.

This is still a kludge, but no more general solution occurs to me offhand.

What other sorts of custom generators are in use? Is there any better way to handle them?

barrysteyn:
Hi Everyone

Just returned from Spain, so I have not looked in detail at this, but I get
the general gist. The fact is, there is no perfect signal to listen for.
Let me settle down (I have literally just walked in) and I'll answer with
something better a little later

avaris:
`all_generators_finalized` is the correct signal, but the issue is mostly how general you expect this to be. 

Expecting a certain structure for *arbitrary* `articles` and/or `pages` attributes is a pretty big assumption, given that the field for custom generators is pretty open and these might be logical internal variables for some other purposes.
 
Honestly, I'd do `isinstance` checks with `pelican.generators.ArticlesGenerator` and `pelican.generators.PagesGenerator` and process `articles` and `pages` for them (respectively). It is safer to assume that the instances from these (or descendants) will behave like expected.

barrysteyn:
Let me test this patch and I will tell you my results. Just to note that the reason why summaries need processing is that cutting off a summary after X characters and destroy Mathjax rendering if there is Math in the summary.

render_math/math.py
```diff
@@ -295,29 +295,38 @@ def pelican_init(pelicanobj):
...
-def parse_articles(article_generator):
-    """Adds mathjax script to RST and processes summaries"""
-
-    for article in article_generator.articles:
-        
-        rst_add_mathjax(article)
-
-        if process_summary.mathjax_script is not None:
-            process_summary(article)
+def massage_content(content_generators):
```
barrysteyn: I don't like the name massage content - it seems too general to me. Also (and this may just be me), the description does not really describe what is being done properly (or it can be described better).

render_math/math.py
```diff
@@ -295,29 +295,38 @@ def pelican_init(pelicanobj):
...
+    """
+    Ensure mathjax is available in content containing math markup.
```
barrysteyn: What do you mean by this. For markdown, mathjax script has already been embedded. So this is only valid for ReStructured Text.

But ReStructured Text will always need to have mathjax appended. Is this the correct place to append it? Answering this questions means answering if this is the correct signal to append any content. Is it general enough and will it work for all generators?

render_math/math.py
```diff
@@ -295,29 +295,38 @@ def pelican_init(pelicanobj):
...
+    
+    We get multiple generators here, and look for textual content,
+    presumed to be in and "articles" or "pages" attribute.
+    That content is decorated with mathjax scripts - including summaries,
+    if it is an article
+    """
```
barrysteyn: The description does not really describe what is being done properly (or it can be described better).

render_math/math.py
barrysteyn: I don't like the name massage_content - rather, this is doing two things:

 (1) Appending mathjax for RST (for markdown, this is already handled)
 (2) Altering summaries if need be.

Rather, change it to parse_content - because rst_add_mathjax is the function that will be altering (massaging) the contents, and same for process_summary.

render_math/math.py
```diff
@@ -295,29 +295,38 @@ def pelican_init(pelicanobj):
...
+    
+    for generator in content_generators:
+        for article in getattr(generator, 'articles', []):
+            rst_add_mathjax(article)
+            #optionally fix truncated formulae in summaries.
+            if process_summary.mathjax_script is not None:
+                process_summary(article)
+        for page in getattr(generator, 'pages', []):
+            rst_add_mathjax(page)
```
barrysteyn: Maybe this has been answered, but how will this work with multiple generators present. I mean, can there be the possibility that summaries and RST will be processed more than once if there are more than one generators present?

render_math/math.py
```diff
@@ -295,29 +295,38 @@ def pelican_init(pelicanobj):
...
```
barrysteyn: Maybe this has been answered, but how will this work with multiple generators present. I mean, can there be the possibility that summaries and RST will be processed more than once if there are more than one generators present?

avaris: For that to happen, same `Article/Page` object has to be present in multiple generators. It doesn't happen in `pelican`, but it is *possible* with a plugin. Though, it's sort of an edge case.

barrysteyn: Okay, I suppose we should just comment somewhere about this. The reason I ask is because I am constantly getting people make contact with me directly about problems which are mostly due to them making cryptic mistakes like this edge case.

avaris: I'm not sure what kind of comment you're thinking but this won't be a mistake users can make. It'll be another plugin doing some specific stuff. Basically, the other plugin has to:

 * create its own generator
 * reuse all or parts of `self.context['articles']` and/or `self.context['pages']` and store them in `self.articles` and/or `self.pages`. (My suggestion regarding `isinstance` will eliminate this to some degree. i.e. it'll add another condition such as the generator has to be a subclass of mentioned `Generator` classes)

But, if you want to be on the safe side, `rst_add_mathjax` and `process_summary` can plant a flag to the `Content` object and check it before doing stuff. This will prevent multiple process of same items.

render_math/math.py
howthebodyworks: I am not deeply attached to any aprticular name of the function so long as it works ;-)

However, I'd recommend not calling it "parse_whatever" because parsing has a particular meaning in computer science generally and in pelican in particular which this function doesn't match. If there is a problem with "massage",  perhaps "process_whatever", "decorate_whatever" etc would be appropriate

render_math/math.py
barrysteyn: I think flagging would be cool, but I wonder if that is going to the extreme (I mean, surely this is a potential problem with all plugins, but not all plugins go to this effort). If you want to flag it, then very cool.

render_math/math.py
barrysteyn: Okay, I agree with parse. Change this to **process_content** - that sounds best

render_math/math.py
barrysteyn: Here is a better description:

  > Ensure Mathjax script is appended to RST files and summaries are processed correctly
  >
  > Iterate through generators, processing only articles and pages (as these
  > assumed to have user content). If article or page has a .rst extension and also
  > also contains math, then insert mathjax script. 
  > 
  > Additionally, if an article and a summary is present, then process the summary
  > if specified by user.

barrysteyn:
Hi Guys

I just altered the PR on my own side. I really don't want to take the credit for the PR, so please can you just look at my repo: https://github.com/barrysteyn/pelican-plugins/blob/master/render_math/math.py

I just renamed a few things and made the comments more sensible. That being said, I will submit the change as a PR if I have not heard anything back by the end of the day.

howthebodyworks:
Hi Barry. Awesome. I'm travelling at the moment and can't do git pushes today.

math.py revised *looks* great. If you leave it until tomorrow I'll incorporate @avaris's suggestion about explicitly checking the type of the generator objects, but if you want the patch to go in right now, I would not object. (I suspect that change will make no effective difference to the users.)

barrysteyn:
Hi

I really don't want to take the cred for this, so I think it can wait till tomorrow. Please can you incorporate my changes that I have in my repo. Doing the check would be great.

justinmayer:
Many thanks, @barrysteyn.

@howthebodyworks: Rebase + squash will also be needed, as can be seen in the many extraneous commits in this PR.  (^_^)

howthebodyworks:
@justinmayer you should see the version in my *private* repo. ;)

The new patch should cleanly apply and perform as expected. I additionally updated the text to fix some typos and formatting issues reflect and to reflect the new functionality.

barrysteyn:
This looks great. @justinmayer, if things are good on your side, I'm happy
to merge this.

justinmayer:
Terrific. Thanks to Dan, Barry, and Deniz for all the effort.  (^_^)

PR #517, opened 2015-06-20
Add always_modified plugin

howthebodyworks:
Say you want to sort by *modified* date/time in a theme template, but not all your articles have modified date/timestamps explicitly defined in article metadata. This plugin facilitates this sorting by assuming the modified date (if undefined) is equal to the created date. _[Description edited as noted below —J.]_

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -36,6 +36,8 @@ Plugin descriptions
 ========================  ===========================================================
 Plugin                    Description
 ========================  ===========================================================
+Always modified           Copy created date metadata into modified date for easy "latest updates" indexes
+
 AsciiDoc reader           Use AsciiDoc to write your posts.
 
 Asset management          Use the Webassets module to manage assets such as CSS and JS files.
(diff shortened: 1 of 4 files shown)


justinmayer:
Original description:

> Say you want to sort by *modified* time, but not all your articles have modified times. This is much easier if article "modified" metadata is copied into the created date field too.

Proposed description:

> Say you want to sort by *modified* date/time in a theme template, but not all your articles have modified date/timestamps explicitly defined in article metadata. This plugin facilitates this sorting by assuming the modified date (if undefined) is equal to the created date.

Do you think that is an accurate assessment of what the plugin does?

justinmayer:
Made some additional in-line comments. Assuming you agree with the proposed changes, please make those modifications and [squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits) down to a single commit. Thanks!

howthebodyworks:
Thanks for the feedback, @justinmayer. I believe the latest pull request incorporates all your suggested changes (up to swapping the word "this" for "that" for clarity in the Readme.md.)

justinmayer:
Thanks for the contribution and the quick response, Dan.

PR #520, opened 2015-06-25
Fixed usage of `iter_commits` in `git-wrapper.py`

vosskuhle:
GitPython renamed the method `commits` of object `Repo` to `iter_commits`. This
has been adressed in commit 47269dd35b9236de7fc84fa99fb9e41c8d975c05

But additionally the attribute `path` has been changed to `paths`. This has
been omitted in the previous commit and is fixed here.

See also [this comment by @a3nm][1].

[1]: https://github.com/getpelican/pelican-plugins/issues/384#issuecomment-68179848


justinmayer:
Thanks for the fix, Michel.

PR #525, opened 2015-07-13
Add missing PagesGenerator import to Summary plugin

makefu:
The summary plugin crashed for me due to missing PagesGenerator import. importing it explicitly Resolves the issue


justinmayer:
Thanks for the fix, Felix!

PR #527, opened 2015-07-18
Always specify parser for Beautiful Soup; fixes extraneous tags

Alexqw:
**This is untested (except for better_figures_and_images, which is how I encountered this bug).**

Beautiful Soup was (silently) using the lxml parser in the better_figures_and_images plugin, and that was resulting in extraneous wrapping with "\<html\>\<body\>" of page.content. According to [the BS docs](http://www.crummy.com/software/BeautifulSoup/bs4/doc/#differences-between-parsers), 'html.parser' is the (non-invasive) parser behavior we want.

Thanks to Avaris and winlu in #pelican for helping track down the root cause of this bug.

---Alex

### DIFF
--- a/better_figures_and_images/better_figures_and_images.py
+++ b/better_figures_and_images/better_figures_and_images.py
@@ -28,7 +28,7 @@ def content_object_init(instance):
 
     if instance._content is not None:
         content = instance._content
-        soup = BeautifulSoup(content)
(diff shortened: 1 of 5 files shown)


Alexqw:
Updated and squashed. Missed slim.py the first time around.

justinmayer:
Thanks for the fix, Alex!

PR #528, opened 2015-07-18
Update plugin to enhance privacy with video embeds

publicus:
Updated the Video Privacy Enhancer plugin, which downloads thumbnails of videos so that users don't get tracking cookies embedded on their systems until they explicitly "opt-in" to watching the video by clicking the thumbnail, at which point the image is changed into the actual iframe video embed. YouTube and Vimeo are now both supported (with the shortcodes `!youtube(video_id)` and `!vimeo(video_id)`, respectively), and the plugin and its documentation are now *much* easier to extend for other video services.


publicus:
I squashed my commits, so if I've formatted everything correctly, just one commit should show up here.

publicus:
Also, in case it helps, a demo of these changes (showing Vimeo support) is [here](http://adunumdatum.org/2015-07-18-video_privacy_enhancer_plugin_updated.html).

justinmayer:
Nice work, @publicus. Thanks for the enhancements!

publicus:
Cheers! : )

PR #530, opened 2015-07-19
Add better-tables pelican plugin

Alexqw:
Simply, this plugin removes the extraneous elements and attributes that rst->html generates. Without this cruft, the tables pass HTML5 validation and make life /much/ easier when styling with CSS.

The real fix is to get rst2html5 merged, but sadly that is beyond both my skillset and time/effort availability.

---Alex

### DIFF
--- a/better_tables/LICENSE
+++ b/better_tables/LICENSE
@@ -1,0 +1,18 @@
+Copyright (c) 2015 Alex Waite
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


justinmayer:
> The real fix is to get rst2html5 merged...

I'm not sure to what this comment refers. Would you please be more specific?

Alexqw:
Ahh yeah. "Merged" is probably the wrong word. I was referring to the abandoned effort to get the rst -> HTML5 generator into pelican. https://github.com/getpelican/pelican/issues/959

My point being that it's be best if the generator just behaved correctly to begin with, rather than going in afterwards and stripping things out.

---Alex

justinmayer:
Thanks for the explanation and the contribution.

PR #532, opened 2015-07-21
New plugin proposal: Emoji

skainz:
Hello,

this is a very simple plugin allowing to render emoji exrpessions e.g. <pre>:+1:</pre> to images, eg.   

:+1: 

 inlined into the article content. Altough this may as well be done using emojify.js, i see no point in having a static weblog renderer needing javascript to display emojis.


justinmayer:
Thanks for the submission, @skainz. Out of curiosity, why images? Why not use Unicode instead?

skainz:
hi,

yes you are right - i will change my code to make it emit unicode
emojis. Thanks for the hint, i just didn't think about that. I'll ping
you again.



Am 2015-07-28 um 18:31 schrieb Justin Mayer:

kylef:
@skainz In that case, why wouldn't you just insert emoji characters directly into your markdown files instead of using this `::` notation?

ulikoehler:
@kylef I see two reasons:
* Because you're used to it from other software (including GitHub)
* Because it's hard to enter special Unicode characters via a keyboard, while entering `:+1:` is easy.

MartinThoma:
@ulikoehler I use http://unicode.party/ if I want to find emoji - pretty easy 🎺🎊🎉😜

PR #533, opened 2015-07-22
Read author from org file metadata

andreas-h:
Read author's name from org-mode file metadata


justinmayer:
Thanks for the enhancement, Andreas!

PR #534, opened 2015-07-25
render_math plugin: force_tls option, prepend protocol to CDN

barrysteyn:
Will not leave cdn protocol as a default value of the document's protocol. Now, user can force tls protocol and document protocol will be forceably pre-pended to CDN.


justinmayer:
Thanks, Barry!

PR #541, opened 2015-08-17
Photos plugin: conserve ICC profiles for more correct colors after scaling/thumbnailing

allanman:


justinmayer:
Thanks for the improvements, @allanman. In the future, you might consider prefixing your commit messages with the plugin name, which would make the commit history easier to read.

Issue #542, opened 2015-08-18
liquid_tags.notebook raises ImportError


terrycojones:
I have python 2.7.6 in a virtualenv on Mac OS X 10.10.3, with ipython 4.0.0 and the master branch of pelican-plugins. In my pelicanconf.py I put `liquid_tags.notebook` in my plugins. When I try to make, I get the following:

```
$ make html
[ ! -d /Users/fred/dark-matter/www/output ] || find /Users/fred/dark-matter/www/output -mindepth 1 -delete
pelican /Users/fred/dark-matter/www/content -o /Users/fred/dark-matter/www/output -s /Users/fred/dark-matter/www/pelicanconf.py
/Users/fred/.virtualenvs/dmp/lib/python2.7/site-packages/IPython/nbconvert.py:13: ShimWarning: The `IPython.nbconvert` package has been deprecated. You should import from ipython_nbconvert instead.
  "You should import from ipython_nbconvert instead.", ShimWarning)
ERROR: Cannot load plugin `liquid_tags.notebook`
  | ImportError: No module named filters.highlight
```

In `pelican-plugins/liquid_tags/notebook.py` both halves of this `if` raise the import error:

``` python
try:
    from IPython.nbconvert.filters.highlight import _pygments_highlight
except ImportError:
    # IPython < 2.0
    from IPython.nbconvert.filters.highlight import _pygment_highlight as _pygments_highlight
```

aavanian:
IPython 4.0 does indeed break this plug-in but strangely, in my situation (Python 3.4 though), it wasn't breaking there (only a warning for this import).

I'm about to do a PR that fixes the one call that breaks the plugin for IPython 4.0

justinmayer:
@terrycojones: Can you try Alexandre's version in pull request #543 and see if that resolves the situation for you?

terrycojones:
This problem went away after I read about the changes in IPython 4.0 and did a `pip install jupyter`. Closing.

terrycojones:
Argh, sorry, I missed the comments here. I'll reopen.

terrycojones:
@justinmayer @aavanian  Yes, thanks, `make html` runs cleanly under the `aavanian:liquid_tags_forward` branch.  Note that the original error went away when I simply modernized my environment for ipython 4. But after doing so I still got an `AttributeError` raised when the code tripped over the `IPython.nbformat` usage. I fixed that by doing

```python
try:
    import nbformat
except ImportError:
[...]

Issue #536, opened 2015-07-28
liquid_tags.notebook: ImportError: cannot import name v1 with ipython-3.2.1


yangj1e:
I tried to use `liquid_tags.notebook` plugin, but with ipython-3.2.1 I got the error:

```
ERROR: Cannot load plugin `liquid_tags.notebook`
  | ImportError: cannot import name v1
```

Does `liquid_tags.notebook` support ipython-3.x?

justinmayer:
Hi @yangj1e. The commit history indicates that the answer to your question is "Yes": https://github.com/getpelican/pelican-plugins/commit/4b4aa408d9e0c8f4c61275e4e06c4788d48763fc

That said, I have no personal experience with that plugin.

aavanian:
Wrong dependancies I think. Use "pip install IPython[notebook]==3.2.1" and that should work.

yangj1e:
It does not work either.

I used ipython 2.4.1 for a while, which works with the plugin. But after recent updates, I got the following error:

```
ERROR: Could not process blog/visualize-hog.md
  | TraitError: No default value found for None trait of <liquid_tags.notebook.SubCell object at 0x109fa90d0>
```

I tried to upgrade IPython to 4.0, but unfortunately I got the same import error as https://github.com/getpelican/pelican-plugins/issues/542

aavanian:
@yangj1e 
May I ask a few more precisions:
* my previous comment, `pip install IPython[notebook]==3.2.1` didn't solve your issue before the PR? after the #543 PR? (though I don't think the PR should affect this outcome)
* under ipython 2.4.1, PR #543 breaks what was working for you previously?
* for ipython 4.0, did you install the dependancies (in particular ipython) this way `IPython[notebook]`?
* Which python are you running (2.7? 3.4?)

@justinmayer 
I was rechecking what I did and realized that somehow, some debug code made it in the PR, I'll send you a new one but would like to see if I can solve this issue first. Apologies for that.

yangj1e:
@aavanian Sorry, my bad. I didn't do `pip install IPython[notebook]` because I'm using zsh and I need to escape the brackets otherwise there is an error, which made me go back to `pip install ipython`.

So `pip install IPython[notebook]` solves the problem, thank you.

aavanian:
No worries! Glad I could help.

PR #543, opened 2015-08-19
Improve ipython forward compatibility (v4.x), also ref #542, #536

aavanian:
This is an attempt to improve compatibility with v4.0 of the IPython / Jupyter packages.

* #536: this is due to dependancies I believe, see amended pip install instruction in the README.md
* #531: this is due to the split of IPython into python specific project (IPython) and notebook and associated utilities (Jupyter). See changes in notebook.py
* I did this by adding a layer of try:/import/except on top of the existing ones. It's brute force but it works though using the correct import after doing a version check might be preferable. Only one change was breaking the module as the nbformat module was called directly instead of being imported so there was no chance the shim provided by IPython 4.0 to ensure backward compatibility. The other changes fix the shim warnings.
* added another change in the REAME.md, a bit unrelated, to avoid the error on the first time the site is built due to the lack of _nb_header.html file. Still have to use the proper read() line depending on Python 2.x or 3.x usage.
* I can't manage to run tox on top of conda environments so please someone run it. I tested it for Python 3.4, IPython 3.2.1 and 4.0.

### DIFF
--- a/liquid_tags/notebook.py
+++ b/liquid_tags/notebook.py
@@ -53,27 +53,51 @@ from .mdx_liquid_tags import LiquidTags
 import IPython
 IPYTHON_VERSION = IPython.version_info[0]
 
+try:
+    import nbformat
+except:
+    pass
+
 if not IPYTHON_VERSION >= 1:
     raise ValueError("IPython version 1.0+ required for notebook tag")
 
 try:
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the valuable contribution, Alexandre. Much appreciated.  (^_^)

PR #544, opened 2015-08-21
Fix and improve extract_toc support for Pandoc reader.

gw0:
Previous commit broke Pandoc reader support, because multiple content readers are capable of handling **same** file extensions. It depends on which ones are enables and on the order which one processes the file.

This fixes the problem and improves detection of which content reader is used. An ideal solution would check exactly which reader class is used, but it seems this information is not available.

### DIFF
--- a/extract_toc/extract_toc.py
+++ b/extract_toc/extract_toc.py
@@ -11,6 +11,11 @@ from os import path
 from bs4 import BeautifulSoup
 from pelican import signals, readers, contents
 
+try:
+    from pandoc_reader import PandocReader
+except ImportError:
+    PandocReader = False
+
 
 def extract_toc(content):
     if isinstance(content, contents.Static):
@@ -20,12 +25,14 @@ def extract_toc(content):
     filename = content.source_path
     extension = path.splitext(filename)[1][1:]
     toc = None
-    # if it is a Markdown file
-    if extension in readers.MarkdownReader.file_extensions:
+
+    # default Markdown reader
+    if not toc and readers.MarkdownReader.enabled and extension in readers.MarkdownReader.file_extensions:
         toc = soup.find('div', class_='toc')
         if toc: toc.extract()
-    # else if it is a reST file
-    elif extension in readers.RstReader.file_extensions:
+
+    # default reStructuredText reader
+    if not toc and readers.RstReader.enabled and extension in readers.RstReader.file_extensions:
         toc = soup.find('div', class_='contents topic')
         if toc: toc.extract()
         if toc:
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the fix, @gw0.

duhoobo:
If I,
* Use `[TOC]` in markdown files,

gw0:
@duhoobo I am not sure what you are trying to do. Are you trying to use `MarkdownReader` and `PandocReader` simultaneously? Why? Have you configured `PandocReader` correctly <https://github.com/getpelican/pelican-plugins/tree/master/extract_toc#pandoc-reader>? Configuration of `PandocReader` is pretty different than for `MarkdownReader`.

duhoobo:
I'm trying to make TOC work with `pelican-elegant` theme (it uses `article.toc` a lot). And all my old articles are all in markdown format.

gw0:
@duhoobo Ok, this just means you need to use the `extract_toc` plugin, but not where is the problem in configuring it correctly? First you need to decide if you want to use the default `MarkdownReader` or the plugin `pandoc_reader`. Then you need to follow instructions for the one you choose (following both instructions simultaneously and partially won't work). In my opinion you should use the default `MarkdownReader` and this means that you disable the `pandoc_reader` plugin (remove it from `PLUGINS = [...]` or even remove the package), because you do not need it for your use case.

duhoobo:
Opening an issue for `pelican-elegant`, its problem!. 

Thanks, anyway.

gw0:
Sorry, but I do not think it is a problem of the theme, but of you not knowing what you want to use and correctly following instructions for what you want. You still haven't answered why do you need to use `pandoc_reader` plugin and do not want to use the default `MarkdownReader`?

duhoobo:
First of all, I definitely want to use the default `MarkdownReader`. After commenting out all `Pandoc` relative code in `extract_toc` in my own working directory, everything just works perfectly as I expected.

Secondly, I don't know much about how `pelican`, `pelican-elegant`, and `extract_toc` are implemented. I don't know how deep I have to dig as a normal user. 

What I have learned is: I write down "[TOC]" in my markdown file, then I need `MD_EXTENSIONS=["toc"]` to let `pelican` generate proper html for it. Then I employ theme `pelican-elegant` and a beautiful toc as sidebar, so I need `PLUGINS=["extract_toc"]`. 

And It just doesn't work. Then I came here to tell you guys what I find out and to seek help. Just, please enlighten me with the correct instructions I should follow, or is what I want to achive just not the right need!

Thanks.

duhoobo:
Besides, I don't have `pandoc_reader` set in `PLUGINS`, and I don't think just removing it from my working copy `pelican-plugins` forcibly is the right way. I must've done something incorrectly, I guess.

gw0:
Thank you for describing the situation more clearly. Unfortunately I do not see what is causing the problem.

Can you confirm that you have the latest version of [extract_toc/extract_toc.py](https://github.com/getpelican/pelican-plugins/blob/master/extract_toc/extract_toc.py) and [pandoc_reader](https://github.com/liob/pandoc_reader/blob/master/pandoc_reader.py)?

Does replacing line 15 (import statement) in `extract_toc/extract_toc.py` with `    pass` also resolve the problem?

Otherwise I found a probably better extension for TOC whose implementation seems less hackish (https://github.com/ingwinlu/pelican-toc). Maybe it is worth checking it out?

PR #550, opened 2015-09-09
Fix non-ASCII code with include_code causing errors

ulikoehler:
(*Edited, now with optional codec selection*)
This pull request solves the long-standing issue that `include_code` does not support files not being written in ASCII. Any UTF8 (or other non-ASCII) character will currently cause an exception and subsequently a pelican crash.

The solution works by defaulting to UTF8 but allowing a codec option in the `include_code` tag:
```
{% include_code NCP380.py lang:python codec:iso-8859-1 %}
```
Basically it is an extension of the solution I proposed in https://github.com/getpelican/pelican-plugins/pull/189. This solution had the problem that non-UTF8 files still could cause a crash.

As UTF8 is a superset of ASCII, this solution guarantees full backwards compability to the ASCII-only `include_code` tag.

### DIFF
--- a/liquid_tags/include_code.py
+++ b/liquid_tags/include_code.py
@@ -6,13 +6,21 @@ based on the octopress video tag [1]_
 
 Syntax
 ------
-{% include_code path/to/code [lang:python] [Title text] %}
+{% include_code path/to/code [lang:python] [Title text] [codec:utf8] %}
 
 The "path to code" is specified relative to the ``code`` subdirectory of
 the content directory  Optionally, this subdirectory can be specified in the
 config file:
 
     CODE_DIR = 'code'
+
+If your input file is not ASCII/UTF-8 encoded, you need to specify the
+appropriate input codec by using the ``codec`` option.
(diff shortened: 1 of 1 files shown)


justinmayer:
Anyone have thoughts on this change? cc: @jakevdp

ulikoehler:
I tested it now and I verified that this solution does not work with ISO-8859-1.

My goal is to introduce an option like that:
```
{% include_code LP2980.py lang:python codec:GB18030 %}
```
with a default of UTF8 (which is 100% backward compatible to the current ASCII default)
but I don't really know how to do that.

ulikoehler:
Actually it wasn't that hard... I believe I fixed it now and once I wrote the documentation for it, I'll edit the description of this pull request.

ulikoehler:
OK, edit is done. The new code should work with all codecs without restrictions.

justinmayer:
Thanks for the enhancement, Uli!

ulikoehler:
Thanks @justinmayer , merging is highly appreciated!

PR #552, opened 2015-09-09
Ability to hide mathjax message and render math in translations

barrysteyn:
Two additions:

 1. User can now control message_style attribute of mathjax
 2. Translations are now also included for rendering math


justinmayer:
Thanks, Barry!

PR #554, opened 2015-09-11
Add Flickr plugin for liquid_tags

xsteadfastx:
The flickr plugin uses the official flickr API to gather all
informations to generate the html tags. The user has to get a official
API key from flickr to use this.

I also added some tests. But i wasnt sure about the state of the tests for liquid_tags. Running tox in that dir tests also now the flickr plugin.


justinmayer:
Thanks for the helpful addition, @xsteadfastx.  (^_^)

PR #555, opened 2015-09-12
Fix parsing of AsciiDoc attribute values by converting to unicode.

jichu4n:
Currently, the AsciiDoc plugin's handling of attributes does not work with list values, such as tags. For example, the unit test currently gives the following error:

```
$ python2 -m asciidoc_reader.test_asciidoc_reader
F.
======================================================================
FAIL: test_article_with_asc_extension (__main__.AsciiDocReaderTest)
----------------------------------------------------------------------
  File "/media/D/Projects/Active/blog/pelican-plugins/asciidoc_reader/test_asciidoc_reader.py", line 47, in test_article_with_asc_extension
    'Actual: %s') % (key, value, page.metadata[key]))
AssertionError: Metadata attribute 'tags' does not match expected value.
Expected: [u'Linux', u'Python', u'Pelican']
Actual: [<Tag 'L'>, <Tag 'i'>, <Tag 'n'>, <Tag 'u'>, <Tag 'x'>, <Tag ','>, <Tag 'P'>, <Tag 'y'>, <Tag 't'>, <Tag 'h'>, <Tag 'o'>, <Tag 'e'>, <Tag 'l'>, <Tag 'c'>, <Tag 'a'>]

----------------------------------------------------------------------
Ran 2 tests in 0.325s

FAILED (failures=1)
```

This is because [ensure_metadata_list](https://github.com/getpelican/pelican/blob/master/pelican/readers.py#L70) in readers.py expects a value of type [six.text_type]( http://pythonhosted.org/six/#six.text_type), which is unicode in Python 2. However, the value returned from the AsciiDoc API is of type str, so it's treated as an iterable directly. Instead of being split on semicolons or commas, the tag attribute is split into individual characters.

This PR fixes this by converting attribute values returned from the AsciiDoc API into six.text_type. The unit test passes after the fix:

```
$ python2 -m asciidoc_reader.test_asciidoc_reader
..
----------------------------------------------------------------------
Ran 2 tests in 0.330s

OK
```

### DIFF
--- a/asciidoc_reader/asciidoc_reader.py
+++ b/asciidoc_reader/asciidoc_reader.py
@@ -10,6 +10,7 @@ File extension should be ``.asc``, ``.adoc``, or ``asciidoc``.
 from pelican.readers import BaseReader
 from pelican.utils import pelican_open
 from pelican import signals
+import six
 
 try:
     # asciidocapi won't import on Py3
@@ -51,7 +52,7 @@ class AsciiDocReader(BaseReader):
         metadata = {}
         for name, value in ad.asciidoc.document.attributes.items():
             name = name.lower()
-            metadata[name] = self.process_metadata(name, value)
+            metadata[name] = self.process_metadata(name, six.text_type(value))
         if 'doctitle' in metadata:
             metadata['title'] = metadata['doctitle']
         return content, metadata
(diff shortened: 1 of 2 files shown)


justinmayer:
Thanks for the fix, Chuan.  (^_^)

PR #410, opened 2015-02-01
Have relevant plugins use all_generators_finalized signal

kernc:
`clean_summary`, `read_more_link`, `representative_image`, `summary` plugins now use `all_generators_finalized` signal so as to avoid issue #314. They fallback to current behavior if said signal is not yet available.

See also: https://github.com/getpelican/pelican/pull/1616

PR #556, opened 2015-09-13
share_post: Reverting removal of summary for the plugin to be usable

### LINKED ISSUE #549: share_post: plugin is unusable ('summary' var non existent)
Plugin won't compile as 'summary' variable on line 42 of share_post.py is not found

davidmarquis:
Otherwise code would not compile (`summary` variable is still used)

Closes #549

### DIFF
--- a/share_post/share_post.py
+++ b/share_post/share_post.py
@@ -27,11 +27,16 @@ def article_url(content):
     return quote(('%s/%s' % (site_url, content.url)).encode('utf-8'))
 
 
+def article_summary(content):
+    return quote(BeautifulSoup(content.summary, 'html.parser').get_text().strip().encode('utf-8'))
+
+
 def share_post(content):
     if isinstance(content, contents.Static):
         return
     title = article_title(content)
     url = article_url(content)
+    summary = article_summary(content)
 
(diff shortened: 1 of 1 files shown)


kernc:
Nitpicking, but revert commits are best made with `git revert` (it's the message that matters).
You could further append a string like

> Closes #549

to the commit message to also auto-close the associated open issue once the change is merged.

(You can change the message of this commit with `git commit --amend` and `git push --force`.)

davidmarquis:
Updated commit and pull request!

justinmayer:
Thanks, @davidmarquis and @kernc, for your contributions here.  (^_^)

avaris:
Umm, won't this bring back the issues related to getpelican/pelican#1616? To be precise, accessing `content.summary` in `content_object_init` signal.

cc: @kernc @davidmarquis

kernc:
Thanks! Indeed it should. :smiley: It was even [part of the same PR](https://github.com/getpelican/pelican-plugins/pull/410/files#diff-edbb62f8c929bbc182e07279a95399c3L30), just didn't contain the signal switch as I probably considered it, without any affected attributes, unnecessary.

Making the plugin fire on `all_generators_finalized` should fix it.

PR #557, opened 2015-09-13
share_post: Remove usage of summary

amitsaha:


justinmayer:
This problem has already been addressed by reverting the relevant change. (See: #556)

amitsaha:
Cool, thanks.

PR #558, opened 2015-09-13
simple_footnotes: modify drafts in addition to articles

lindzey:


justinmayer:
Thanks for the enhancement, Laura!  (^_^)

PR #559, opened 2015-09-14
liquid_tags: Add soundcloud plugin

xsteadfastx:
The soundcloud plugin uses the official soundcloud API to get the widget
html tag. Also added tests.


justinmayer:
Thanks for the nice enhancement, @xsteadfastx!

PR #562, opened 2015-09-18
liquid_tags: Add audio and giphy plugins

xsteadfastx:
The Audio plugin allows using the html5 audio tag to embedded audio files to posts. its inspired by the video plugin.

The Giphy plugin uses the official API to get gif informations and creating a img tag to add gifs to posts.


justinmayer:
Thanks for the enhancements, @xsteadfastx!

PR #569, opened 2015-09-27
headerid: Make anchor target a configurable option

albertodonato:
This adds support for configuring which char to use for headers anchor targets.


justinmayer:
Thanks for the enhancement, Alberto!  (^_^)

PR #570, opened 2015-09-29
Relative import for code_include

fufler:
This PR makes code_include plugin compatible with python3 import statement syntax.


justinmayer:
Thanks, Alexey.

fufler:
np :)

PR #571, opened 2015-09-29
liquid_tags: Fix alignment in {% graphviz %} output

Xion:
In liquid_tags plugin, the `graphviz` tags produces incorrect HTML that doesn't align the diagram as intended. This PR fixes it.


justinmayer:
Thanks for the fix, Karol!

PR #572, opened 2015-09-29
Update pelican-jinja2content and pelican-md-metayaml

joachimneu:
Open for your feedback. :)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -52,9 +52,9 @@
 [submodule "pelican-jinja2content"]
 	path = pelican-jinja2content
 	url = https://github.com/joachimneu/pelican-jinja2content.git
-[submodule "pelican-md-yaml"]
-	path = pelican-md-yaml
-	url = https://github.com/joachimneu/pelican-md-yaml.git
+[submodule "pelican-md-metayaml"]
+	path = pelican-md-metayaml
+	url = https://github.com/joachimneu/pelican-md-metayaml.git
 [submodule "panorama"]
 	path = panorama
 	url = git://github.com/romainx/panorama.git
--- a/code_include/__init__.py
+++ b/code_include/__init__.py
@@ -1,1 +1,1 @@
-from code_include import *
+from .code_include import *


justinmayer:
I renamed `pelican-md-yaml` to `md-metayaml` via 8a0d6d8. Will make the other changes shortly.

joachimneu:
Just one question: Do you intend to remove the "pelican-" prefix on all plugins eventually, or was that accidental? (Now the plugin's readme states `"pelican-md-metayaml"` as plugin name to be added to `PLUGINS`, which is true, if the plugin is cloned from its repo into a `plugins` folder, but false, if the plugin is installed from `pelican-plugins`. I'd like to avoid this inconsistency and confusions likely resulting from it.)

justinmayer:
Intentional. I find the `pelican-` prefix redundant and unsightly, and I've made an effort to remove that prefix wherever possible.

kdeldycke:
The new convention around pelican plugin names and repository IDs is now enforced as part of the general migration process out of the mono-repo.

This PR can be considered as outdated and safely closed.

justinmayer:
@kdeldycke: Agreed. Thanks for reviewing.

I wonder how many Pelican plugins exist that handle YAML front-matter metadata. If there are multiple, it would be nice to consolidate them into a single repo in https://github.com/pelican-plugins, perhaps basing it on the https://github.com/joachimneu/pelican-md-metayaml plugin repo, assuming @joachimneu is amenable, of course. 😄

kdeldycke:
@justinmayer I should probably add the YAML plugins to my general overlap survay at: https://github.com/getpelican/pelican-plugins/issues/1295

justinmayer:
@kdeldycke: That would be great. While you're adding the YAML metadata plugins, you can include this one, too: 😊 https://github.com/noirbizarre/pelican-frontmark

logological:
> I wonder how many Pelican plugins exist that handle YAML front-matter metadata. If there are multiple, it would be nice to consolidate them into a single repo in https://github.com/pelican-plugins, perhaps basing it on the https://github.com/joachimneu/pelican-md-metayaml plugin repo, assuming @joachimneu is amenable, of course. smile

pelican-md-metayaml doesn't work with recent versions of Pelican, and despite [reports of this problem](https://github.com/joachimneu/pelican-md-metayaml/issues/12) going back two years, the plugin hasn't been updated.  However, I did find another YAML plugin for Pelican at https://github.com/pR0Ps/pelican-yaml-metadata which seems to do everything that pelican-md-metayaml does, and moreover works with Pelican 4.x. This might also be relevant for Issue #1295.

PR #576, opened 2015-10-01
Use https instead http to get secure resources content in blog over h…

bameda:
Use https instead http to get secure resources content in blogs run over https.


kylef:
Thanks @bameda

bameda:
:dancer: you are welcome

PR #579, opened 2015-10-04
Add commas indicating tuple to example

Moini:
Added commas to indicate a tuple in the example. 
Without them, loading of known author avatar will fail (at least it did for me).


Scheirle:
Yes there should be commas, thanks.

Moini:
(Sorry for asking: I'm not sure about the 'workflow' now - do I have to do anything?)

justinmayer:
@Moini: You don't have to do anything; I'll take it from here. Thanks for the fix!

Moini:
Thanks!

PR #580, opened 2015-10-05
liquid_tags: gram plugin Python 3 compatibility

xsteadfastx:
There were a few things to change. I changed the import to get it
working under python3 for urlopen and used items() instead of iteritems().


justinmayer:
Thanks for the enhancement!

PR #583, opened 2015-10-12
MAINT: add warning for IPython versions greater than 1.x

jakevdp:
This is mainly so people will stop asking me why the plugin doesn't work :smile:


jakevdp:
Long-term it would be nice to update the plugin, but conflicting CSS schemes are a real pain. I tried making it work with IPython 2.0 shortly after it was released, but gave up after working for several hours and getting nowhere.

justinmayer:
Many thanks, Jake.

PR #585, opened 2015-10-13
Update notebook.py

thongrim:
Fixed a typo in the if IPYTHON_VERSION > 1 which was raising a NameError: name 'IPython_VERSION' is not defined


thongrim:
Already requested in #584

PR #586, opened 2015-10-14
add post_revision plugin

jhshi:
Add post revision plugin, which enables you to show post revision history from Git commits.

See a live example here:

http://jhshi.me/2015/10/13/post-revision-plugin-for-pelican/index.html


justinmayer:
Added via 5200565. Thanks for the contribution, Jinghao!

PR #588, opened 2015-10-14
Fix 87ce0a7

sjpfenninger:


justinmayer:
Thanks for the contribution. However, this has already been addressed via https://github.com/getpelican/pelican-plugins/commit/570e8e4e7a0a71bdb3f7250b2cdc475784363336. (^_^)

PR #589, opened 2015-10-15
Add ga_page_view plugin

jhshi:
This plugin pulls the page view information from Google Analytics and add the page view count as a meta of article or page.

See a live example here:

http://jhshi.me/

Note that each post has its own page view count.


justinmayer:
Thanks for the contribution, Jinghao!

PR #593, opened 2015-10-17
Add Sub parts to plugin list. Fix Markdown.

pxquim:
My plugins were missing from Readme.rst. This patch fixes both that and deviant Markdown syntax.


pxquim:
Oh No! Arrgghhh!

PR #599, opened 2015-10-29
better_figures_and_images: img_path is redundant and breaks plugin

xenithorb:
markdown expects {filename}/images/file.jpg format, with leading forward slash. This is exemplified by the fact that if you try to rectify this bug bug leaving off the leading slashing, the following error is produced:

```
> /usr/lib/python2.7/site-packages/pelican/generators.py(587)generate_output()
-> self.generate_pages(writer)
(Pdb) n
WARNING: Unable to find `images/dnf_changelog.png`, skipping url replacement.
```

Presumably we only care about candidates if they're in a proper static path, so img_path is redundant and will produce results like `images/images`, which then leads to `if no src_candidates` being false.

### DIFF
--- a/better_figures_and_images/better_figures_and_images.py
+++ b/better_figures_and_images/better_figures_and_images.py
@@ -57,7 +57,7 @@ def content_object_init(instance):
                 # 2. Build the source image filename from STATIC_PATHS
                 src = os.path.join(instance.settings['PATH'], img_path, img_filename)
                 src_candidates = [src]
-                src_candidates += [os.path.join(instance.settings['PATH'], static_path, img_path, img_filename) for static_path in instance.settings['STATIC_PATHS']]
+                src_candidates += [os.path.join(instance.settings['PATH'], static_path, img_filename) for static_path in instance.settings['STATIC_PATHS']]
                 src_candidates = [f for f in src_candidates if path.isfile(f) and access(f, R_OK)]
 
                 if not src_candidates:


if1live:
If i use static image, I use this code.
``fixed font``
``../``  is attached.

But In pelican document, ``../`` is used for document, not static file.
sorry. this bug is caused by my enviroment. I use wrong syntax.


In my document style,
`{filename}../static/sample/sidetail-sora.gif`

```python
path = '/home/haruna/devel/libsora.so/content'
image_path = '../static/sample'
image_filename = 'sidetail-sora.gif'

os.path.join(path, image_path, image_filename)
# /home/haruna/devel/libsora.so/content/../static/sample/sidetail-sora.gif
```

In your document style,
`{filename}/static/sample/sidetail-sora.gif`

```python

path = '/home/sora/devel/libsora.so/content'
image_path = '/static/sample'
image_filename = 'sidetail-sora.gif'

os.path.join(path, image_path, image_filename)
# /static/sample/sidetail-sora.gif
# lost path information 
```

so I suggest hotfix code.

Remove first '/' to fix os.path.join works.
https://github.com/if1live/pelican-plugins/commit/3ea69436576ba15ea045d4c04b9bb180a29f05f0#diff-ff6ec3f224c260f9f2fe7d44021517c1R60

these are related codes.
```
# original code
src = instance.settings['PATH'] + img_path + '/' + img_filename

# buggy code
src = os.path.join(instance.settings['PATH'], img_path, img_filename)

# alternative code
img_path = os.path.sep.join([el for el in img_path.split("/") if len(el) > 0])
src = os.path.join(instance.settings['PATH'], img_path, img_filename)
```

I think this code works well your environment. Can you test it? @xenithorb

xenithorb:
if1live@3ea69436576ba15ea045d4c04b9bb180a29f05f0 Tested as working for me. Thanks

PR #600, opened 2015-10-30
Fixes for pelican-cite plugin

cmacmackin:
Now removes any unescaped curly braces from the BibTeX output. This looks better visually and also avoids a bug whereby curly braces can confuse Python's `format` methods. I also fixed a bug where the names of book authors were being shortened. Finally, I added documentation on the ability to specify a different BibTeX file for each article. This feature had previously existed, but I had forgotten to document it.


justinmayer:
Thanks, Chris!

PR #614, opened 2015-11-20
Make pdf-related settings not required

trovao:
rst2pdf can work without the settings the plugin referenced.


justinmayer:
Thanks, @trovao!

Issue #573, opened 2015-09-30
pelican-plugins/pdf/ extension to md


deguss:
dear developer @kylewm ! Is there a way to get working the pdf generator on markdown files also? Many thanks!

kylewm:
Hi @deguss, unfortunately I didn't write the PDF extension -- I just refactored it out of the main repository into a separate plugin a couple years ago during a code sprint.

The plugin itself delegates the actual conversion to an rst2pdf library. If there's a comparable markdown-to-pdf library, it should be easy to adapt, but I'm not seeing any great candidates in a quick search...

deguss:
thank you! I am also not able to find any usable tool for converting md to pdf. I should rather search for html2pdf.

trovao:
Hi @deguss, you might want to check whether this new update to the PDF plugin works for you.

PS: Sorry about the pull request spam.

justinmayer:
Not sure if this is still a problem. If so, please feel free to comment here so the issue can be re-opened.

PR #616, opened 2015-11-22
RFC: Add support for exporting Markdown articles as PDF. Refs #573

trovao:
I'm not sure this is the best code we could have. If you think its style is a bit too much, please let me know.

This may be considered a hack, but it actually builds upon a documented feature
of rst2pdf: you can use raw HTML in your RST files and rst2pdf will parse them.

Therefore, the idea of this commit is:

 1- Load the markdown source files and the markdown reader;
 2- Convert the markdown file to an HTML representation;
 3- Massage the HTML a bit so it can be considered valid RST;
 4- Convert the newly-created RST to PDF.

For this feature to work, the use must have xhtml2pdf installed. A bug in
rst2pdf (https://github.com/rst2pdf/rst2pdf/issues/521) prevents it from
parsing raw HTML correctly, so we monkeypatch xhtml2pdf to provide the class
rst2pdf expects.

### DIFF
--- a/pdf/pdf.py
+++ b/pdf/pdf.py
@@ -10,17 +10,36 @@ from __future__ import unicode_literals, print_function
 
 from pelican import signals
 from pelican.generators import Generator
-from rst2pdf.createpdf import RstToPdf
+from pelican.readers import MarkdownReader
 
 import os
 import logging
 
+try:
+    from markdown import Markdown
+except ImportError:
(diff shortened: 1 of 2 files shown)


trovao:
I've found a leftover from a previous experiment in the commit referenced in this pull request. I'm closing it and creating a new one.

PR #617, opened 2015-11-23
RFC: Add support for exporting Markdown articles as PDF. Refs #573

trovao:
This is the newest iteration of my previous improvement to the PDF plugin.

In my local tests, Markdown articles & pages are converted just fine. With the exception of template pages, which I have no idea how to convert.

Another drawback is that absolute image references aren't handled automatically.


trovao:
Just noticed the latest commit invalidates this patch series / pull request. So, since this is not going to apply nicely, I'm closing it.

PR #620, opened 2015-11-26
New plugin: section numbering

potatola:
For some themes, the relationship between different sections can be vague. This plugin add section number to the section title. To make it looks like below:

# 1 bla
## 1.2 bla
# 2 bla


justinmayer:
Thanks for the contribution, @potatola!

PR #621, opened 2015-11-28
Tag cloud with badges

mothsART:
Hi , I don't found an easy way to add a "badge" ( bootstrap terminology ) on my cloud of tags.
like this
tags_with_badges

That's why i propose this patch separate on 2 commits :
* With extension "**TAG_CLOUD_BADGE**" option
* Unit tests to cover before

I cleaned the two python files (**tag_cloud.py** and **test_tag_cloud.py**) after a **pep8** and **pyflake** checking...

I hope it will serve to other.

Best regards, Jérémie Ferry

### DIFF
--- a/tag_cloud/tag_cloud.py
+++ b/tag_cloud/tag_cloud.py
@@ -7,26 +7,30 @@ This plugin generates a tag cloud from available tags
 from __future__ import unicode_literals
 
 from collections import defaultdict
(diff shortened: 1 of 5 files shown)


avaris:
Can you document the new setting in the readme file?

mothsART:
Avaris : Done!

PR #623, opened 2015-11-30
Change Panorama plugin submodule URL scheme from git to https

MarcoForte:
Changed panorama git submodule url from git to https. In an effort to be able to use pelican-plugins as a submodule. 
Before when attempting to setup a github pages blog I got the error,   
> The submodule registered for `./panorama` could not be cloned. Make sure it's using https:// and that it's a public repo. For more information, see https://help.github.com/articles/page-build-failed-invalid-submodule.
  
So I made the panorama submodule url public, tested it and it worked for me.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -54,7 +54,7 @@
 	url = https://github.com/joachimneu/pelican-jinja2content.git
 [submodule "panorama"]
 	path = panorama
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks, Marco!

PR #626, opened 2015-12-10
Gravatar plugin > Standardize quotation marks

dakerfp:


justinmayer:
Not sure how this is "standardizing"; the existing style seems to be consistent with [generally accepted practice](http://stackoverflow.com/a/56190).

dakerfp:
It is consistent with the practice, but is not consistent with itself: 
metadata["author_gravatar"]
metadata['email']

justinmayer:
You are quite right -- somehow I read it backwards, as if you were making the change in the other direction. #notenoughcoffee

Thanks for the fix!

dakerfp:
You're welcome.

PR #632, opened 2015-12-21
Python 3 support for pelican-cite, figure-ref, and pdf-img plugins

cmacmackin:
Also a bug fix for pelican-cite. Getting rid of curly braces in the citation label had caused a bug resulting in the letter 'a' being appended to the year in the label. This is now fixed.


justinmayer:
Thanks, Chris!

PR #451, opened 2015-03-31
New test process #449 + test case on liquid tags notebook

benjaminabel:
This PR shows the advantages of the new test process described in #449 fixing issues(#445, #427) due to recent update of ipython to python3 with backward incompatible changes.
More work needs to be done in liquid tags templates ,styles, and js, however it **fixes the generation and test it**.

Issue #449, opened 2015-03-29
Improve test suite


benjaminabel:
Currently tests are passed with nose, and you can't test your plugin against multiple requirements or configurations. It would be nicer to let each plugin maintainer to choose the way tests are run against his plugin.
I've created a fork in which I try to implement a new way to pass tests you can see it [here](https://github.com/benjaminabel/pelican-plugins/commit/7acf63f204d6eb5afb15e5fca4cad828c66e5078). **It's  a work in progess and needs your feedbacks and reviews**

Here are the main ideas:

- A `runtests.py` file has been added to the root folder. It contains a list of plugins which needs or wants to pass the test suite. It then run the tests for each selected plugin.
- travis only calls the `runtests.py` file.
- if a package maintainer wants to add his tests to the test run, just add the plugin name to the `PLUGINS` list in `runtests.py`, and then add a `runtest.py` script to run the tests suite in the plugin folder.

**Linting**
In the example provided the flake8 linting is run at the plugin level, but it may be runned in the selected plugins at the root level to improve coding styles between plugins.

In advance, thanks for your feedbacks.

benjaminabel:
Updated in #636

PR #636, opened 2015-12-26
Improve testing

benjaminabel:
A  `runtests.py` script has been created in the project main folder that:
- check for tests in every plugin
- run tests for each plugin with a default tox.ini file or tox if a tox.ini file is present
- write results to a csv file [tests_results.csv](https://github.com/benjaminabel/pelican-plugins/blob/improve-testing/tests_results.csv)

_Possible improvements_
- Currently the csv file is written locally but it would be better to write it from travis after each build.
- Run a default generation test for each plugin that don't implements any test(lmay be very long!)
- Add coverage report to the csv file + latest commit date in the plugin dir to see if plugin is maintained.
- Add the build svg status logo.

@justinmayer here the updated test process described in #449 as requested in discussion #451. It only implements the process with the liquid_tags plugin but other can be added easily. You can see the build log on travis [here](https://travis-ci.org/benjaminabel/pelican-plugins)


justinmayer:
Nice work, @benjaminabel!

@avaris / @ingwinlu: Any thoughts on these changes?

PR #637, opened 2015-12-29
Org reader metadata

andreas-h:
support reading most relevant metadata from org header properties.

any further properties we should include?

**CAUTION:** this is backwards-incompatible: previously, the `CATEGORY` property was translated into *tags*, while now `CATEGORY` is translated into category, and for tags there's `PROPERTY: tags`.

I don't really like the `#+PROPERTY: key value` syntax here; however my elisp is at a very early stage, and I couldn't find out how to extract `#+TAGS: `, `#+MODIFIED: ` and the like directly.

Any improvements to elisp style and such are more than welcome =)

@ardekantur @dsteuer @zakkak

### DIFF
--- a/org_reader/org_reader.py
+++ b/org_reader/org_reader.py
@@ -2,7 +2,7 @@
 Org Reader
 ==========
 
-Version 1.0.
+Version 1.1.
 
 Relevant Pelican settings:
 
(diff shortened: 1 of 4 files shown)


org_reader/org_reader.py
```diff
@@ -16,13 +16,28 @@
...
   to 'html.
 
-To provide metadata to Pelican, provide the following header in your Org file:
+To provide metadata to Pelican, the following properties can be defined in
+the org file's header:
 
 #+TITLE: The Title Of This BlogPost
 #+DATE: 2001-01-01
-#+CATEGORY: comma, separated, list, of, tags
+#+CATEGORY: blog-category
```
zakkak: Maybe add in parentheses that it can be a comma separated list

zakkak: Sorry, my wrong, Pelican seems to not support multiple categories per article

zakkak:
Looks good to me. Thanks @andreas-h !

One improvement I can think of, is the addition of error handling, i.e., when `#+TITLE:` is not set it would be nice to print a descriptive error message.  Currently it gives the counter-intuitive `Wrong type argument: stringp, nil`.

zakkak:
Another possible improvement would be to add the start of the article as SUMMARY when the latter is not set as an org property.  This will also be closer to the default behavior of markdown processing. The change is straight forward, but If you need any help, I can submit a pull request.

andreas-h:
good suggestion.  would i search for a new paragraph (how to reliably 
detect this?), or rather just take the first N characters of the post? 
If the latter, what would be a good measure of N?  Can you tell me how 
the markdown reader does it?

zakkak:
TBH I have no clue what's the actual approach of the markdown processing.  Inspecting some of the articles on my website and their auto-generated summary it seems to get the first 50 words of the post.

andreas-h:
Actually, the solution is easy: Just don't always set the `summary` 
metadata to an empty string in the python code ;)  Fixed that now.

andreas-h:
Once you guys give me the go-ahead, I'll squash this PR into one single commit; before that I'll leave it as is for better readability.

dsteuer:
Thanks a lot for your work! Looks like quite an improvement to me, so go ahead!

zakkak:
Just tested the summary change as well and looks great.  Thanks again! Please go-ahead and merge this upstream :) :+1:

andreas-h:
okay, from my part, this is ready to be merged upstream

justinmayer:
Thanks for the enhancements, Andreas. And thanks to @zakkak and @dsteuer for reviewing!

PR #641, opened 2016-01-03
Add mboxreader plugin as submodule

TC01:
This is a plugin I wrote a while back but just recently cleaned up, ran through PEP8, submitted to PyPI and this repository.

Given a path to a Unix mbox, it implements a custom generator that runs over all emails in that mbox and turns them into Pelican articles. The main utility for me was to turn a Mailman archive into blog posts (we had an announcements email list that I wanted automatically mirrored to a Pelican site), but I could vaguely see other uses (some kind of auto submission system?).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -118,3 +118,6 @@
 [submodule "ace_editor"]
 	path = ace_editor
 	url = https://github.com/mothsART/ace_editor.git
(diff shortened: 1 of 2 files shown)


justinmayer:
Thanks for the contribution, Ben!

PR #644, opened 2016-01-13
Summary: add SUMMARY_USE_FIRST_PARAGRAPH

Schnouki:
Makes it possible to only use the first paragraph of a post as its summary if no marker is found.


justinmayer:
Thanks for the enhancements, Thomas!

Issue #461, opened 2015-04-11
org_reader: "Symbol's function definition is void: org-export-get-environment"


andreas-h:
I get the error *Symbol's function definition is void: org-export-get-environment* when trying to build a simple website containing of one `index.org` file.  I have tried both without setting `ORG_READER_EMACS_SETTINGS`, and also with pointing `ORG_READER_EMACS_SETTINGS` to a file with the contents

    ; Enable org-mode
    (add-to-list 'load-path "/home2/hilboll/src/org-mode/org-8.2.10/lisp")
    (require 'org)

I'm sure I'm missing something simple, but would be glad if you could point me to a solution.

andreas-h:
got it, I was missing a `(require 'ox)`.

PR #647, opened 2016-01-17
Org_reader plugin: fix org-export-get-environment is void

GeiliCode:
The error below occurs when I do "make publish".
```
Symbol's function definition is void: org-export-get-environment
ERROR: Could not process ./2016-01-aerosolve-demo-image-impressionism.org
  | CalledProcessError: Command '[u'/usr/local/bin/emacs', '-Q', '--batch', '-l', '/home/zhang/blog/plugins/org_reader/org_reader.el', '--eval', '(org->pelican "/home/zhang/blog/src/content/2016-01-aerosolve-demo-image-impressionism.org" \'html)']' returned non-zero exit status 255
```

I found the solution in https://github.com/getpelican/pelican-plugins/issues/461, and it works.

My envirment:
GNU Emacs 24.5.1
pelican 3.6.3

### DIFF
--- a/org_reader/org_reader.el
+++ b/org_reader/org_reader.el
@@ -1,5 +1,6 @@
 (require 'json)
 (require 'org)
+(require 'ox)
 (defun org->pelican (filename backend)
   (progn
     (save-excursion


justinmayer:
Thanks to @GeiliCode and @andreas-h for your work on fixing this!

PR #649, opened 2016-01-18
rmd_reader: import knitr library via rpy2.importr

wilsonfreitas:
Instead of passing code to R some rpy2 functions have been used to keep the
code clear.
Rpy2 warnings have been suppressed.


justinmayer:
Thanks, Wilson!

PR #652, opened 2016-01-23
org_reader: move language's setting form `#+LANGUAGE lang` to `#+PROPERTY: LANGUAGE lang`

GeiliCode:
If `#+LANGUAGE` is not in the ORG file, the default value is not None nor Empty, it will be the language of OS. This will change the filename to `{slug}-{language}.html`, and the ARTICLE_SAVE_AS in `publishconf.py` will not be used.


justinmayer:
Thanks, Andrew!

PR #653, opened 2016-01-25
Photos: Handle SITEURL

### LINKED ISSUE #607: photos-plugin: Photos in article are using wrong hard-coded URL instead of SITEURL
When including an photo as image in an (.rst-)article as

     .. image:: {photo}name.jpg

It is correctly recognized and resized. The image src-URL however is wrongly translated.

In photos.py in detect_content the line

                origin = os.path.join('/photos', photo)

hard-codes the location without respecting SITEURL. This works when the blog is located in the top directory on the server (e.g. when running the server with "make serve"). When the blog is located in a subdirectory or when the photos are located on another location, this fails.

Therefore I suggest to introduce an option PHOTO_BASEURL and to change the above line to

                origin = os.path.join(settings['PHOTO_BASEURL'], photo)

with PHOTO_BASEURL = SITEURL as default when not set in the pelicansettings.py.

The galleries are not affected by this bug since their URLs are encoded in the Jinja-Templates which include SITEURL in the appropriate place.

pxquim:
Fixes #607. Articles with embedded images now work when SITEURL has a path.

### DIFF
--- a/photos/photos.py
+++ b/photos/photos.py
@@ -105,7 +105,7 @@ def detect_content(content):
                 logger.error('photos: No photo %s', path)
             else:
                 photo = os.path.splitext(value)[0].lower() + 'a.jpg'
-                origin = os.path.join('/photos', photo)
+                origin = os.path.join(settings['SITEURL'], 'photos', photo)
                 enqueue_resize(
                     path,
                     os.path.join('photos', photo),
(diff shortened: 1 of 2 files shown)


justinmayer:
If you're only relying on the domain and not the existence of the specified image at that domain, then I think the `getpelican.com` domain should be fine. Thanks!

PR #655, opened 2016-01-25
Add support for Knitr configuration options

aaronk:
I modified the knitr setup in the rmd_reader plugin so that I could pass options to knitr_opt_chunk via a pelicanconf option KNITR_OPTS_CHUNK, e.g.

KNITR_OPTS_CHUNK = {
    'fig.path': 'images/', 
}

Mostly I wanted to set fig.path, but it should work for a number of the other options that would make sense.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.pyc
 *.log
 *~
+.project
+.pydevproject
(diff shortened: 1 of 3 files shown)


justinmayer:
Nice work, Aaron. Thanks for the enhancement!

PR #662, opened 2016-02-02
Add Linker plugin

JuliusR:
This plugin allows to define custom linker commands in analogy to the builtin `{filename}`, `{attach}`, `{category}`, `{tag}`, `{author}`, and `{index}` syntax.

By default, it supports one additional linker command: `{mailto}`.


justinmayer:
Thanks for the contribution, @JuliusR!

PR #663, opened 2016-02-03
filetime_from_git: Fix bad function call

Tblue:
A really simple patch: `datetime_from_timestamp()` is not a function provided by the Git wrapper,
it's a module-level function of the calling module.


justinmayer:
Thanks for the fix, Tilman.

PR #664, opened 2016-02-03
filetime_from_git: Use relative imports

Tblue:
The filetime_from_git plugin needs to use relative imports; otherwise, it doesn't seem to work using Python 3.


justinmayer:
Thanks for the fix, Tilman.

PR #666, opened 2016-02-03
filetime_from_git: Add GIT_FILETIME_ALWAYS_ADD_MODIFIED setting.

Tblue:
The plugin usually always sets the "modified" field of a content piece,
even if it has not actually been modified, leading to the "Modified" date being the same as the "Created" date in some cases. This setting allows the user to disable that behaviour so that only content pieces which have been modified get a "Modified" date.

To ensure backward compatibility, this setting is set to "True" by default -- i. e. the old behaviour is kept by default.


Tblue:
Then again, maybe the old behaviour should simply be considered a bug -- any thoughts?

PR #669, opened 2016-02-08
Add pelican-resume plugin.

cmenguy:
Pelican module to generate customizable PDF resumes based on Markdown pages.
Currently only supports moderncv resume style, based on the CSS from http://resume.chmd.fr/style.css but other resume styles can be added by writing a custom CSS and the style can be controlled by the `RESUME_STYLE` setting.


bittner:
Please rebase to make it ready for merging!

PR #673, opened 2016-02-15
Update for submodule post_revision

maweki:
This updates the post_revision plugin which adds python3 support.


maweki:
Is there any problem with merging this PR? What is the prefered way to get a submodule updated?

justinmayer:
This is the correct method. Sincere apologies for the delay!

PR #677, opened 2016-02-18
[pelican_comment_system] Increase interoperability with other plugins

Scheirle:
- readme: Updates urls
 - The main logic starts now with the `article_generator_finalized` signal.
   This allows other plugins to access the comment information of the articles.
 - uses the active writer (maybe replaced by other plugins) to write feeds (Don't recreate the default one multiple times)


justinmayer:
Thanks, @Scheirle!

PR #679, opened 2016-02-20
Added support for footer link in the summary

okelet:
If new variable FEED_USE_SUMMARY_FOOTER set in config, a footer in the RSS summary will be added; default content is:

<br/>The post <a href="{posturl}">{posttitle}</a> was first published in <a href="{siteurl}">{sitename}</a> by {authorname}.

But can be customized using variable FEED_USE_SUMMARY_FOOTER_CUSTOM.


okelet:
Already using it in https://okelet.github.io/feeds/all.atom.xml.

PR #675, opened 2016-02-17
Update org_reader's README.md

GeiliCode:
If you use the default path of emacs `/usr/bin/emacs` as `ORG_READER_EMACS_LOCATION`, org_reader will not find relative packages, because the packages of `Emacs for Mac OS X` are installed in the folder `/Applications/Emacs.app/Contents`.

PR #680, opened 2016-02-22
Update org_reader's README.md

GeiliCode:
If you use the default path of emacs `/usr/bin/emacs` as `ORG_READER_EMACS_LOCATION`, org_reader will not find relative packages, because the packages of `Emacs for Mac OS X` are installed in the folder `/Applications/Emacs.app/Contents`.


GeiliCode:
Derived from #675.

PR #686, opened 2016-03-05
[pelican_comment_system] Adds Quickstart guide + default theme

Scheirle:
* Now it should be easier to setup the system (including the comment form)
* No functional changes


justinmayer:
Thanks, @Scheirle!

PR #687, opened 2016-03-07
Add Yuicompressor plugin as submodule

auroredea:


justinmayer:
Thanks for the submission, Aurore.

PR #690, opened 2016-03-18
Add filetime_from_hg plugin

douardda:
this is mostly an adapted copy of filetime_from_git to target mercurial
repositories.


justinmayer:
Thanks for the addition, David.

PR #691, opened 2016-03-19
Allow user to change mathjax urls in render_math

chaserhkj:
Here in China our network condition is poor, sometimes connection to mathjax cdns may stuck and prevent the equations from loading.

Thus I enabled changing of `source` settings in `render_math` plugin, allowing user to change urls to their preferred CDNs or mirrors in case of bad network condition.


justinmayer:
Thanks for the enhancement, Kang-Jing.

PR #692, opened 2016-03-23
render_math: clarify scope of macro support

howthebodyworks:
I would like to introduce LaTeX macros support (e.g. for `\renewcommand` and the limited macro definitions that MathJaX allows)

However, I need to update the documentation first because otherwise when discussing math display environments it seems to imply that some kind of user macro support already exists.

This patch clarifies that we are discussing supporting display math environment macros in particular, rather than arbitrary macros (which in the broad sense are not in any case supported by all our rendering backends)

### DIFF
--- a/render_math/Readme.md
+++ b/render_math/Readme.md
@@ -129,17 +129,20 @@ must *not* be any whitespace before the ending `$`. So for example:
  * **Will not render as inline math**: `$40 vs $50`
 
 ###Displayed Math
-Math between `$$`..`$$`, for example, `$$`x^2`$$`, will be rendered centered in a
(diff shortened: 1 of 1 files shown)


howthebodyworks:
@barrysteyn may be keen to be involved in this one.

barrysteyn:
Speaking of macro support, I applied a PR to my personal repo that enhances the macro support. See https://github.com/barrysteyn/pelican_plugin-render_math/pull/27

I would love someone to test and look at it over. @howthebodyworks If you could give it a look over and play round with it, I would appreciate your feedback.

barrysteyn:
@howthebodyworks Your change looks good. I am happy with it.

justinmayer:
Thanks, Dan and Barry!

PR #693, opened 2016-03-23
[render_math] Mathjax templated javascript

danmackinlay:
This pull request updates render_math to generate the mathjax script tags using jinja templates instead of python string templating.

This is a largeish change and it would be great if we had unit tests to ensure it hadn't broken anything ;-)

However, this should make it easier to write code with less bugs in the future; The string-interpolation method requires javascript-strings-in-javascript-strings-in-python-strings in the configuration file, which is difficult to write when it works, and doesn't work if your string needs to include quotation marks.

Jinja templating should ease development of new settings for the render_math plugin by eliminating these problems, and since jinja is used by pelican anyway should hopefully not add substantially to load time to memory usage.

Additionally, while implementing this I updated the templates to produce 2 script tags for Mathjax - a configuration script tag and a mathjax script tag. Combining these two is deprecated by Mathjax.


howthebodyworks:
Whoops, I forgot to squash that down! Anyway, let's see if there's any comment first.

howthebodyworks:
And I notice now that I've failed to re-enable HTTP access to mathjax in this code as opposed to HTTPS/TLS. Although - what is the use-case for this? Who will be desperate to access Mathjax over HTTP? A user with a private install of Mathjax on their intranet might wish to do this - but we currently don't support custom MathJax URLs. Other uses? I'd recommend we deprecate that feature in any case, or supplement it with custom Mathjax URL support.

howthebodyworks:
@barrysteyn might have an opinion on this one, and in particular [has a branch with LaTeX macro support which may relate to this one](https://github.com/barrysteyn/pelican_plugin-render_math/pull/27)

justinmayer:
@barrysteyn: Any thoughts on this pull request?

barrysteyn:
@justinmayer I am on leave from work for a bit, so I will look into it within the next two or three days...

barrysteyn:
Hi - this is quite a big change. I know I was guilty of not making tests, however I believe a change like this needs to have some tests in order to see if things will work. @howthebodyworks perhaps we can collaborate on some tests?

howthebodyworks:
My enthusiasm for tests is strictly proportional to the number of users of the code minus one. ;-)

Sure, I'll try to squeeze in tests as well; I'm currently travelling and will be for the next month so this may take a while.

In the interim, I should report that i am  blogging using my own fork including the above pull request with the jinja templating and it Works For Me (tm)

howthebodyworks:
Wow; Looking back, since I render_math I've done a lot of work over the year on my render_math branch, including fixing handing of math markup in drafts and adding support for Pelican 3.7. Untangling this mess could take a while.

justinmayer:
@howthebodyworks: I know it's probably a pain, but your work on this is greatly appreciated! (^_^)

howthebodyworks:
Right back atcha @justinmayer.
The timing is nonetheless unfortunate - I *had* time earlier.

Also there is a moving target issue here - see e.g. #787 #805...

danmackinlay:
I think that based on the current projections, the chance of me writing tests (#829) for render_math in any finite time are low, as it is at this point are low down a very long TODO list full of things for my day-job, including managing a boss who thinks that this new "Web" thing is a transitory fad.

Instead, I've moved on and created what is effectively substantive rewrite of render_math, with some new features and somewhat less code. This is not ideal, but it keeps my blog running. It does not, however, produce a nice clean pull request.

PR #694, opened 2016-03-31
Replace YUICompressor submodule with actual plugin

auroredea:
I had problems with python path.
Deleted submodule and changed name.


justinmayer:
Wonderful. Sub-modules are a pain. Thanks, Aurore!

PR #695, opened 2016-04-01
Sitemap plugin: Parse template_pages and write them in sitemap

ziogaschr:
As a developer I would like my template pages to be written in sitemap.xml

NOTE: I initially thought that Pelican will parse template pages as contents.Page, but it seems that's not true.

Let me know if you prefer me to implement this in a different way.


ziogaschr:
Sorry guys I forked the initial repo from the wrong repo. I will close this PR and create a new one.

PR #696, opened 2016-04-01
Sitemap plugin: Parse template_pages and write them in sitemap

ziogaschr:
As a developer I would like my template pages to be written in sitemap.xml

NOTE: I initially thought that Pelican will parse template pages as contents.Page, but it seems that's not true.

Let me know if you prefer me to implement this in a different way.

### DIFF
--- a/sitemap/sitemap.py
+++ b/sitemap/sitemap.py
@@ -239,6 +239,19 @@ class SitemapGenerator(object):
                                 save_as=standard_page_url)
                 self.write_url(fake, fd)
 
+            # add template pages
+            for path, template_page_url in self.context['TEMPLATE_PAGES'].iteritems():
+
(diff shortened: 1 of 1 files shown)


mitchtbaum:
How well would this approach suit your needs?
https://github.com/getpelican/pelican/wiki/Tips-n-Tricks#generate-sitemapxml

ziogaschr:
I tried this and it works pretty well for me. Although I thought that other might need TEMPLATE_PAGES to be part of the plugin.

Let me know if you think it's better to not merge it in the plugin and I will close this PR.

Thanks

justinmayer:
Seems like a useful addition to the plugin. Folks can always choose which approach suits them best.

Thanks for the contribution, Chris!

PR #697, opened 2016-04-03
Add pelican-version plugin as submodule

Shaked:
# Pelican Version Generator

A simple version generator which generates an incremented version file.

# Usage

```
PLUGINS = [
    'pelican_version'
]

VERSION_PATH = PATH + '/version'
OUTPUT_VERSION_PATH = 'output'

```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -124,3 +124,6 @@
 [submodule "bootstrap-rst"]
 	path = bootstrap-rst
 	url = https://github.com/Alexqw/bootstrap-rst
+[submodule "pelican-version"]
+	path = pelican-version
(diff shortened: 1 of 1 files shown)


justinmayer:
Out of curiosity, what is the envisioned use case for this plugin?

Shaked:
I am using it for deployment. I never deploy by deleting old versions,
instead I keep the version number so in case something happens I can always
relink it to the previous version.


__________________
Shaked Klein Orbach
http://www.shakedos.com/about

justinmayer:
Thanks for the explanation and the contribution.  (^_^)

PR #700, opened 2016-04-05
rmd_reader: Use relpath sans ext of markdown source as prefix for figures to avoid overwriting figures with same chunk name

jar1karp:
This commit changes the behaviour of RMD_READER_RENAME_PLOT variable such that it can be set to chunklabel to enable the default unnamed chunk renaming, or to directory to rename figures according to the path of the markdown source. The latter is intended to be used in conjuction with setting fig.path in knitr options. Alternatively the variable can be set to anything else to disable automatic renaming, in which case the chunks in markdown files must have unique names.


justinmayer:
Hi Jari. Thanks for the proposed enhancement. It might be a good idea to show a helpful error message if someone has the variable set to the previous `True`/`False` Boolean setting, no?

jar1karp:
Agree, also I had forgotten to update the unit test. Those two new commits should fix both

justinmayer:
Wonderful. Thanks, Jari!

PR #702, opened 2016-04-12
rmd_reader: Fix for python3

jar1karp:
Python3 doesn't have has_key() for dict anymore, should use key in dict instead.


jar1karp:
If using python3 you will get the error

AttributeError: 'dict' object has no attribute 'has_key'

with the recent changes to rmd_reader. This commit fixes the error and works for both python2 and python3.

justinmayer:
Thanks for the fix, Jari, and also very the clear description of what the fix addresses. That's always super helpful!  (^_^)

PR #704, opened 2016-04-16
Make events plugin work with i18n_subsites

stieglma:
Previously the events plugin and i18n_subsites could not be used together. Two issues had to be fixed:
1. Due to rewriting the output-path to a not already existing folder of i18n_subsites the events plugin failed with an IOError during writing the .ics file. Now we check the existance of the path before writing.

2. For a multi-language site we do also want to have multi-language calendars and not one calendar with every language in it (and thus, multiple events at the same time with differently named events although they are the same). To fix this I added some checks and when i18n_subsite plugin usage is detected we filter out duplicates of the events and create the calendar file on a per language base and not for the overall folder.

### DIFF
--- a/events/events.py
+++ b/events/events.py
@@ -118,8 +118,24 @@ def generate_ical_file(generator):
     ical.add('prodid', '-//My calendar product//mxm.dk//')
     ical.add('version', '2.0')
 
-    for e in events:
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the compatibility fix, Thomas!

PR #706, opened 2016-04-18
asciidoc_reader: Don't add empty metadata values

sabaini:
The asciidoc api will return unset document attributes as empty strings. When set as metadata this will cause errors. This workarounds by pruning empty values before setting them as metadata.


justinmayer:
Thanks for the enhancement, Peter!

PR #707, opened 2016-04-21
liquid_tags.youtube: change regex to support width & height in percents

snay:
The liquid_tags youtube module will support both absolute and percent size of video player.

Now it's only possible to do: {% youtube VIDEO_ID 640 390 %}
Will be possible to do so as well: {% youtube VIDEO_ID 640% 390% %}


justinmayer:
Thank you for the enhancement, @snay.

PR #708, opened 2016-05-02
related_posts: Also fill article list for draft articles

ionelmc:
n/t


justinmayer:
Thanks for the enhancement, Ionel!

PR #710, opened 2016-05-03
Fixes for Events plugin

stieglma:
In pull request #704 I already fixed some parts of the events plugin which did not work in combination with i18n_subsites. It turned out that there were even more issues and my commits did also not work with Python 3. This pull-request should fix this. In detail:
- Events were shown multiple times on the homepage (generated ical was correct, that was the first PR)
- Python 3 compatibility restored

### DIFF
--- a/events/events.py
+++ b/events/events.py
@@ -31,6 +31,7 @@ TIME_MULTIPLIERS = {
 }
 
 events = []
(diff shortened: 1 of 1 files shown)


justinmayer:

PR #711, opened 2016-05-03
Photos plugin enhancement

stieglma:
The photos plugin works nicely for monolingual homepages. Together with i18n_subsites the galleries are processed once for each language, and an additional time for the initial page generation. This is of course not what should happen. The pictures are the same for a certain gallery independant of the language of the post which uses the gallery.

With this pull request I add some memoization to the photos plugin such that the galleries are only processed once and not multiple times.

Additionally I added the possibility of specifying multiple galleries for one post. This is as easy as putting them in the same :gallery: tag e.g. ```:gallery: {photo}folder1 {photo}folder2```
Please note that this feature does only work with the ```{photo}``` syntax and not with ```{filename}```

### DIFF
--- a/photos/photos.py
+++ b/photos/photos.py
@@ -132,38 +132,51 @@ def detect_content(content):
 def process_gallery_photo(generator, article, gallery):
     if gallery.startswith('/'):
         gallery = gallery[1:]
-    dir_gallery = os.path.join(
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the enhancements, Thomas.

PR #713, opened 2016-05-15
pelican_comment_system: Correct template path

konrad:
Just a minor typo in the docs.


Scheirle:
Nice catch. :+1:

Please also rename the `template` folder¹ to `templates` to make it consistent.

¹: `pelican_comment_system/theme/template`

konrad:
Good point. Done.

Issue #714, opened 2016-05-15
PELICAN_COMMENT_SYSTEM_DIR requires absolute path


konrad:
I have set up the pluging "pelican_comment_system" and created the folder "comments" in the pelican directory next to "content", etc and added dummy comments as suggested by the docs. With the default setting of PELICAN_COMMENT_SYSTEM_DIR (="comments") or when setting it explicitly to "comments" the comments are not processed.

```
pelican content
Processed 0 comment(s)
Done: Processed 2 articles, 0 drafts, 0 pages and 0 hidden pages in 0.14 seconds.
```
Only when I set the variable to the absolute path they seem to be processed (actually comments still don't occur in the post but this seems to be another issue).

```
pelican content
Processed 2 comment(s)
Done: Processed 2 articles, 0 drafts, 0 pages and 0 hidden pages in 0.15 seconds.
```

I use the current version (3.6.3) and just cloned the current version of the plugin-repo.

Scheirle:
As mentioned in the documentation `PELICAN_COMMENT_SYSTEM_DIR` is relative to `PATH`.
I assume¹ your `PATH` variable points to your `content` folder and therefore your `comments` folder should be inside your `content` folder. (This location is totally fine, as the comment system ensures that the comments are not processed as articles or pages)

¹: it's the default if you created the site with the `pelican-quickstart` command.

> actually comments still don't occur in the post but this seems to be another issue

I suspect a error in the theme. Is your site source public?

konrad:
Thanks, @Scheirle, your are right. I wrongly assumed (just starting to play around with pelican) that `PATH` is the folder that contains the `content` folder as subfolder while the `pelicanconf.py`clearly sets  `PATH` as `content`. Setting the `PELICAN_COMMENT_SYSTEM_DIR` to `../comments` fixes my problem so this is not a bug but my mistake.

> > actually comments still don't occur in the post but this seems to be another issue

> I suspect a error in the theme. Is your site source public?

Thanks for offering to have a look at that. I generated a repo with the test site.
https://github.com/konrad/pelican_test

Scheirle:
Nope wasn't a theme error.

1. In `1.md` you have `replyto: `. As there is no comment with an empty (` `) slug, `1.md` has no parent comment and therefore isn't visible.
2. The pelican slug method does remove dots so in `2.md` you should replace `replyto: 1.md` with `replyto: 1md`.

---

I'll add a warning if we processes a comment with an invalid parent.
[...]

PR #715, opened 2016-05-15
[pelican_comment_system] Logs a warning if the parent of a comment can not be found.

Scheirle:
https://github.com/getpelican/pelican-plugins/issues/714#issuecomment-219302129 :
> I'll add a warning if we processes a comment with an invalid parent.
> So in the future such errors are easy to detect.


justinmayer:
Thanks, @Scheirle!

PR #717, opened 2016-05-17
Fix filetime_from_git not working in python3

Aetf:
Fix filetime_from_git not working in python3 when `GIT_FILETIME_FOLLOW = True`. This fix uses list comprehension instead of map which works in both python2 and python3.


justinmayer:
Thanks for the fix, @Aetf!

PR #720, opened 2016-05-21
[pelican_comment_system] Fix mailto link: use '\r\n' instead of '\n'

### LINKED ISSUE #719: [pelican_comment_system] Creation of mailto url should follow the mailto protocol RFC
The RFC: http://www.faqs.org/rfcs/rfc2368.html

Specifies about encoding in Section 5: 
"Also note that line breaks in the body of a message MUST be encoded with "%0D%0A".

Currently pcs only uses %0A ( 0A == \n ).

Scheirle:
Fixes #719


justinmayer:
Thanks for the fixes, @Scheirle!

PR #729, opened 2016-06-05
extract_toc: Fix pandoc_reader issue

### LINKED ISSUE #582: Question: extract_toc doesn't work if pandoc isn't available but pandoc_reader is
I've been trying to use the extract_toc plugin and the pelican_comment_system together, but when I include both into the plugins list, I keep getting a 'File not found' error: 

Traceback: https://gist.github.com/Moini/feb9946a93a1a6282a7e
Installed eggs: https://gist.github.com/Moini/b12dc3e60dfaa9146542
Settings: https://gist.github.com/Moini/ba41742e641760b68c5e

When I take one of them out, it works again. My Pelican version is still 3.5.0 (I'm trying to do one thing at a time ;) ) I've tested with a non-custom theme (simple) and get the same error. 

Update: With pelican 3.6.3, the errors are different, but the result is the same (no website generated):
https://gist.github.com/Moini/7c88a5a6cddf5ed0ff5c

Please let me know if you need more files/info, or if I can try out something that might shed light onto the issue.

bnice5000:
Fixes #582

toc_extract assumes that if it can import pandoc_reader that
pandoc_reader is configured. This is not always the case and can cause
an error as pandoc_reader requires pelicanonf parameter.

This fix adds a plugin check to make sure that pandoc_reader was meant
to be called by having it as a configured setting.

### DIFF
--- a/extract_toc/extract_toc.py
+++ b/extract_toc/extract_toc.py
@@ -10,18 +10,16 @@ place it in its own `article.toc` variable for use in templates.
 from os import path
 from bs4 import BeautifulSoup
 from pelican import signals, readers, contents
+import logging
 
-try:
-    from pandoc_reader import PandocReader
-except ImportError:
-    PandocReader = False
+logger = logging.getLogger(__name__)
 
 
 def extract_toc(content):
     if isinstance(content, contents.Static):
         return
 
-    soup = BeautifulSoup(content._content,'html.parser')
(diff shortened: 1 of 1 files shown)


ShaunaGordon:
:+1:

justinmayer:
Thanks for the fix, Brian!

PR #733, opened 2016-06-17
[Photos] Blacklist support

fufler:
This commit provides a way to specify a list of blacklisted images that should be ignored when processing gallery folder. If `blacklist.txt` file is present in gallery folder then it's treated as a list of files that should be ignored. There are a lot of ways to create such a file (for example parsing tags in EXIF), but it's up to user to decide how to generate it.

### DIFF
--- a/photos/photos.py
+++ b/photos/photos.py
@@ -44,6 +44,14 @@ def read_notes(filename, msg=None):
             logger.warning(msg, filename)
     return notes
 
+def read_blacklist(filename):
+    notes = {}
+    try:
+        with pelican_open(filename) as text:
+            return set(text.split())
+    except:
+        return set()
+
 
 def enqueue_resize(orig, resized, spec=(640, 480, 80)):
     global queue_resize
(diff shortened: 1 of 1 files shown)


bnice5000:
@fufler I rolled your changes into my refactoring of photos. Let me know if it meets your specs.

fufler:
@bnice5000 It's ok, but I would suggest to update plugin documentations since it's not obvious which one is blacklisted with following config file 

```
# best.jpg
new.jpg
```

I suppose the commented out line should be deleted to avoid any misunderstanding.

bnice5000:
User documentation is not my strong suit. I would love to have some additional help with the documentation if you are willing.

Also, trying to go from just ok to great. I was working off your implementation. If I have messed up the feature let me know. I would gladly accept help making it conform more closely to your implementation.

fufler:
Your implementation is fine. I'll send another pull request with more verbose `blacklist.txt` example.

PR #734, opened 2016-06-17
Fix for documentation based on changes to photos.

### LINKED ISSUE #723: [Photos] Current documentation does not work
I just set up the Photos Plugin with my blog. When trying to adjust the template as stated in the README file pelican produed an error due to the fact that
article.photo_gallery does not, as mentioned, hold 5-tuples for each photo inside the gallery associated to the article but a tuple of the gallery name in its first element and a list with the 5-tuples describing the pictures in its second element.

I changed the code added to the article.html template file as follows:
```
{% if article.photo_gallery %}
  <div class="gallery">
    {% for gallery in article.photo_gallery %}
      {% for name, photo, thumb, exif, caption in gallery[1] %}
        <a href="{{ SITEURL }}/{{ photo }}" title="{{ name }}" exif="{{ exif }}" caption="{{ caption }}"><img src="{{ SITEURL }}/{{ thumb }}"></a>
      {% endfor %}
    {% endfor %}
  </div>
{% endif %}
```
As I am not yet very familiar with Pelican, Plugins, and theming, this solution works but does not feel to be the correct one. I am sure that there is a more correct way of doing it.

Anyway, the documentation should probably be corrected to reflect that problem.

bnice5000:
Fixes #723 based on changes made in commit #272fcb57d11b662880059177ec810fa0e375771e

### DIFF
--- a/photos/README.md
+++ b/photos/README.md
@@ -89,8 +89,10 @@ For example, add the following to the template `article.html` to add the gallery
 
 		{% if article.photo_gallery %}
 		<div class="gallery">
-			{% for name, photo, thumb, exif, caption in article.photo_gallery %}
-			<a href="{{ SITEURL }}/{{ photo }}" title="{{ name }}" exif="{{ exif }}" caption="{{ caption }}"><img src="{{ SITEURL }}/{{ thumb }}"></a>
+			{% for title, gallery in article.photo_gallery %}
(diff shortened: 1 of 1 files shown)


bnice5000:
Incorporated this PR into a larger refactor of the plugin. See PR #749

PR #735, opened 2016-06-19
[Better Figures and Images] Major Rework

bnice5000:
I want to start by saying that I am making a pull request to start a discussion. This is a complete rework of your plugin. While it has many similar features. It also removes some features and adds others. 

Specifically, I removed Pillow and its work getting the image size. I did this because it breaks support for the Photos plugin.

I then add support for Bootstrap and I add some Markdown specific support for additional tag attributes and converting an img tag to a figure tag.

I completely understand if you think this should be a separate plugin. However, I wanted to submit it as a pull request first because I started with your plugin as the base code. It was also my inspiration for the idea.

Thank you for your consideration of these changes. I look forward to hearing your thoughts.


bnice5000:
Not the best fit. There is a plugin that already does some of what I was working to do with this plugin. Moved the code to a fork of that plugin.

PR #736, opened 2016-06-19
Sync all submodules to latest commits

kdeldycke:


justinmayer:
Thanks, Kevin!

kdeldycke:
Thanks @justinmayer ! That was fast! :)

PR #752, opened 2016-07-13
Add read time feature

deepakrb:
This plugin calculates the read time for each article based on the number of words and a words per minute count. This is quite commonly seen on medium articles. (See [here](https://www.deepakrb.com/blog/) for how it looks in pelican). There are two ways to set up the plugin: using words per minute only or using the more advanced language option.


Scheirle:
There is already a plugin which does something quite similar: https://github.com/getpelican/pelican-plugins/tree/master/post_stats

Maybe both plugins should reference each other.

deepakrb:
@Scheirle Well damn, I definitely could have saved myself some time here

PR #749, opened 2016-07-06
photos: Refactor of photos, bug fixes, additional features

bnice5000:
Fixes #723 
Fixes #546
Incorporated pull request #733 
Incorporated discussion on commit 272fcb5

This plugin also adds the ability to add watermarks to images.

PR #754, opened 2016-07-20
Photos: Fix Python 3 incompatibility

bnice5000:
Per comment to PR #749 by @fufler. Here is the patch to fix python 3 compatibility.

@justinmayer I have not tested it as I do not have a python 3 environment. I can try to set one up and give it a go later this evening.

### DIFF
--- a/photos/photos.py
+++ b/photos/photos.py
@@ -6,6 +6,7 @@ import logging
 import os
 import pprint
 import re
+import sys
 
 from pelican.generators import ArticlesGenerator
 from pelican.generators import PagesGenerator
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks, Brian. If possible, please do check any and all contributions with Python 3. It's fairly easy to use Virtualenv to create a `pelican-py3` virtual environment. (See [related docs](http://docs.getpelican.com/en/stable/contribute.html#setting-up-the-development-environment).)

Also, in the future please prefix the commit message with the name of the plugin; otherwise it's hard to follow the commit history. (e.g., `Photos: fix Python 3 compatibility`)

PR #757, opened 2016-07-23
[better_figures_and_images] Default to 100% width if opening image with PIL fails (e.g. for SVG images)

jar1karp:
The better_figures_and_images plugin stops with an error if it fails to open any image to parse its width. For example, this would happen with RMarkdown articles with SVG plots. Instead, according to [this article](http://tympanus.net/codrops/2014/08/19/making-svgs-responsive-with-css/) we could just set the width to 100%.


justinmayer:
Thanks for the enhancement, Jari.

PR #758, opened 2016-07-23
Summary: clean div tags

gbraekmans:
There are some open div tags in the summary if you're using rst and headings. This will remove these tags from the summary.


justinmayer:

PR #759, opened 2016-07-24
rmd_reader: set initialization options for R environment

wilsonfreitas:
The initialization options: --no-save, --quiet and --vanilla, have been set to create an clean R environment ignoring specific user settings.
In the early created R environment the locale has been set to "C" to avoid i18n problems while parsing data.

### DIFF
--- a/rmd_reader/rmd_reader.py
+++ b/rmd_reader/rmd_reader.py
@@ -19,6 +19,8 @@ def initsignal(pelicanobj):
     try:
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
+            import rpy2.rinterface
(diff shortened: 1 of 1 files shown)


jar1karp:
76654fd seems to break `make devserver` -- I'm getting `RuntimeError: Options cannot be set once R has been initialized.` with rpy2 2.8.3.

justinmayer:
@wilsonfreitas: Thanks for the enhancement. Do you want to squash any of these commits or merge as-is?

wilsonfreitas:
Hi @jar1karp, this last commit solves the problem.
Thanks for pointing.

wilsonfreitas:
@justinmayer please merge as-is.
Thanks.

PR #762, opened 2016-07-31
Add Glossary plugin for definition lists found in articles and pages

leotrs:
By creating the `definitions` variable, visible to all page templates, which contains all information from definition lists (<dl> tags) found in articles.


justinmayer:
Thanks for the contribution, Leo. Nice work!  (^_^)

PR #767, opened 2016-08-10
liquid_tags: added speakerdeck tag

TheNeuralBit:
Added a speakerdeck tag, as seen here:
https://github.com/TheNeuralBit/theneuralbit/blob/master/content/2014-07-19-viterbi-analysis.md

http://theneuralbit.com/viterbi-analysis.html


justinmayer:

PR #768, opened 2016-08-14
Make representative_image Py3 compatible

kernc:


justinmayer:
Thanks for the enhancement, @kernc!

PR #770, opened 2016-08-18
Photos: Add ability to keep EXIF data on resize

bnice5000:
Added exif copy. Refactored jpeg rotate. Both now rely on piexif.
Added tool to keep exif if the original input photo was a jpg.
Added tool to remove gps data from jpeg exif.
Added debug option to disable multiprocess
Added tool to add copyright and artist information to jpeg exif.
Added tool to set default color on alpha channel images when they are converted to jpg.
Cleaned up README.txt.

### DIFF
--- a/photos/photos.py
+++ b/photos/photos.py
@@ -1,13 +1,15 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import datetime
 import itertools
+import json
 import logging
+import multiprocessing
 import os
 import pprint
 import re
 import sys
-import multiprocessing
 
 from pelican.generators import ArticlesGenerator
 from pelican.generators import PagesGenerator
@@ -17,15 +19,22 @@ from pelican.utils import pelican_open
 
 logger = logging.getLogger(__name__)
 
-
 try:
-    from PIL import ExifTags
     from PIL import Image
     from PIL import ImageDraw
     from PIL import ImageEnhance
     from PIL import ImageFont
 except ImportError:
-    raise Exception('PIL/Pillow not found')
+    logger.error('PIL/Pillow not found')
+
+try:
+    import piexif
+except ImportError:
+    ispiexif = False
+    logger.warning('piexif not found! Cannot use exif manipulation features')
+else:
+    ispiexif = True
+    logger.debug('piexif found.')
 
 
 def initialized(pelican):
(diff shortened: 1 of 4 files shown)


photos/photos.py
```diff
@@ -155,56 +184,121 @@ def watermark_photo(image, watermark_text, watermark_image, watermark_image_size
...
 
-def resize_worker(orig, resized, spec, wm, wm_text, wm_img, wm_img_size):
+def rotate_image(img, exif_dict):
+
+    if "exif" in img.info and piexif.ImageIFD.Orientation in exif_dict["0th"]:
+        orientation = exif_dict["0th"].pop(piexif.ImageIFD.Orientation)
+        if orientation == 2:
+            img = img.transpose(Image.FLIP_LEFT_RIGHT)
+        elif orientation == 3:
+            img = img.rotate(180)
```
allanman: we could (should?) use transpose() with Image.ROTATION_180 etc  instead of rotate(). 

reason being I had an issue where rotate did no longer rotate the dimensions of the image and I got black bars+cropping as a result (maybe i updated Pillow too far?)

bnice5000: The `transpose` and `rotate` perform the same operations. The only difference is the transpose operation implicitly sets the `expand` parameter to true. 

> The transpose(ROTATE) operations can also be performed identically with rotate() operations, provided the expand flag is true, to provide for the same changes to the image’s size.

http://pillow.readthedocs.io/en/3.3.x/handbook/tutorial.html#transposing-an-image
http://pillow.readthedocs.io/en/3.3.x/reference/Image.html
http://pillow.readthedocs.io/en/3.3.x/reference/Image.html#PIL.Image.Image.rotate

The code it replaces used the same operation to rotate the image 180.

```python
for tag, value in exif.items():
                    decoded = ExifTags.TAGS.get(tag, tag)
                    if decoded == 'Orientation':
                        if   value == 3: im = im.rotate(180)
                        elif value == 6: im = im.rotate(270)
                        elif value == 8: im = im.rotate(90)
                        break
```
https://github.com/bnice5000/pelican-plugins/blob/712607b3fc6190516b8278e2eebdc4d0c4a4f07e/photos/photos.py

The code I have take to replace it is a reproduction of the piexif example code.
http://piexif.readthedocs.io/en/latest/sample.html#rotate-image-by-exif-orientation

The past behavior and the documented behavior use the same code. We could change it, but the implicit expand could have unintended consequences.

justinmayer:

PR #778, opened 2016-09-14
Rename knitr object to KNITR for consistency

### LINKED ISSUE #776: rmd_reader: global name 'knitr' is not defined
In `rmd_reader.py` knitr options are set by using `knitr` object. However, it is imported into the `KNITR` object. This results in a `NameError` when used for parsing.

jar1karp:
This should fix  #776


justinmayer:
Thanks for the fix, Jari.  (^_^)

PR #779, opened 2016-09-15
better_figures_and_images: Add SVG support

ohporter:
Add SVG support to better_figures_and_images using pySVG.


justinmayer:
Thanks for the contribution, Matt.

Any comments on this, @dflock?

justinmayer:
Thanks for the review, Duncan!

ohporter:
@dflock Indeed, only the <object> tag is handled which was the case I encountered including some interactive SVG CPU FlameGraphs. I wish I knew more about HTML to do a better job but I'm just a kernel developer trying to do project docs. :) Thanks for the review and the excellent plugin!

Issue #488, opened 2015-05-11
Multiple titles occur in asciidoc file 


lijunsong:
Reproduce the problem:

name a file `foo.asciidoc`, which contains the following text:

```
:title: TITLE
:date: 2013-3-4


Paragraph1

Paragraph2

Paragraph3

Paragraph4
```

The output body is

```
<div class="paragraph"><div class="title">TITLE</div><p>Paragraph1</p></div>
<div class="paragraph"><div class="title">TITLE</div><p>Paragraph2</p></div>
<div class="paragraph"><div class="title">TITLE</div><p>Paragraph3</p></div>
<div class="paragraph"><div class="title">TITLE</div><p>Paragraph4</p></div>
```

Note the repeated `TITLE`.

marco-m:
As another data point, this still happens in September 2016 with pelican 3.6.3, a current asciidoc plugin and asciidoc 8.6.9.
In my opinion this makes asciidoc unusable with Pelican.

justinmayer:
Thanks for the data point, Marco. I don't believe any of the maintainers use Asciidoc, so we could use some help. Would you be willing to assist us track down the problem and address it?

avaris:
To be honest, this is an issue with `asciidoc`. If you run that file with `asciidoc` directly, you'll see that the generated html has that in the content. Unless `asciidoc` itself is fixed, I don't see anything we can do here.

marco-m:
I did some more direct tests (only asciidoc processor without pelican).

- Running `asciidoc` gives the duplicated titles, as @avaris pointed out.
- Running `asciidoctor` (a more recent Python re-implementation of asciidoc) gives the expected output (no duplicated titles).

I also found a workaround: changing the input file as follows: remove the `:title:` attribute and use the explicit asciidoc syntax for the title as follows:

```
= TITLE
:date: 2013-3-4

Paragraph1

Paragraph2

Paragraph3

Paragraph4
```

will generate the correct output with pelican. This is good enough for me. Thanks both @justinmayer and @avaris.

I would suggest closing this issue as resolved. I will write a PR for the pelican documentation with a sample asciidoc article in the same spirit as the existing rST and Markdown examples.

marco-m:
PR: https://github.com/getpelican/pelican/pull/2013

PR #783, opened 2016-09-19
Add AsciiDoc metadata format example

marco-m:
Following discussion for issue 488 of pelican-plugins (getpelican/pelican-plugins#488)


justinmayer:
Thanks for the contribution, Marco!

PR #785, opened 2016-09-23
filetime_from_git: better py2 and py3 compatibility

jar1karp:
Python 3 doesn't have izip_longest introduced in commit f5d897632977abd0ef9ea591347331e60063ffac -- apparently it's called zip_longest. For compatilibity with py2 and py3 this commit just implements the cookbook idiom (alternative 1) from 
http://python-future.org/compatible_idioms.html#itertools-filterfalse-zip-longest

Downside is that this requires the future package to be installed i.e. `pip install future`.

### DIFF
--- a/filetime_from_git/git_wrapper.py
+++ b/filetime_from_git/git_wrapper.py
@@ -2,7 +2,10 @@
 """
 Wrap python git interface for compatibility with older/newer version
 """
-import itertools
+try:
+    from itertools import zip_longest
+except ImportError:
+    from six.moves import zip_longest
 import logging
 import os
 from time import mktime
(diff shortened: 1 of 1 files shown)


filetime_from_git/git_wrapper.py
```diff
@@ -2,7 +2,7 @@
 """
 Wrap python git interface for compatibility with older/newer version
 """
-import itertools
+from future.moves.itertools import zip_longest
```
justinmayer: Since the [`six` module is already a dependency](https://github.com/getpelican/pelican/blob/master/setup.py#L8), it's probably better to use that method. Also, using try/except blocks here might be better. Something like:

    try:
        from itertools import zip_longest
    except ImportError:
        from six.moves import zip_longest

jar1karp: Good points, I've updated the fix accordingly. Also verified that it works with both versions.

justinmayer:
Nice work. Thanks for the fix! (^_^)

PR #786, opened 2016-09-25
convert <div> to <span> on a liquid_tag_video

mothsART:
Liquid tags wrap content in a **<p>** tag.
Unfortunaly, the **<p>** element represents a paragraph. It cannot contain block-level elements like **<div>** (https://www.w3.org/TR/html401/struct/text.html#edef-P)

All the browser trying to interpret this so

``` html
<p></p>
<div>
    ... my content
</div>
<p></p>
```

I suggest to replace **<div>** by a **<span>** tag.

### DIFF
--- a/liquid_tags/video.py
+++ b/liquid_tags/video.py
@@ -14,9 +14,11 @@ Example
 
 Output
 ------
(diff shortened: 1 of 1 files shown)


justinmayer:

PR #787, opened 2016-09-25
render_math: Fix for new MARKDOWN dict in Pelican 3.7

Krastanov:
This extensions does not work with the current git version of pelican. This fix hopefully addresses the issue (works for me).

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -269,7 +269,10 @@ def mathjax_for_markdown(pelicanobj, mathjax_script, mathjax_settings):
 
     # Instantiate markdown extension and append it to the current extensions
     try:
-        pelicanobj.settings['MD_EXTENSIONS'].append(PelicanMathJaxExtension(config))
+        if isinstance(pelicanobj.settings.get('MD_EXTENSIONS'), list): # pelican 3.6.3 and earlier
+            pelicanobj.settings['MD_EXTENSIONS'].append(PelicanMathJaxExtension(config))
+        else:
+            pelicanobj.settings['MARKDOWN'].setdefault('extensions', []).append(PelicanMathJaxExtension(config))
     except:
         sys.excepthook(*sys.exc_info())
         sys.stderr.write("\nError - the pelican mathjax markdown extension failed to configure. MathJax is non-functional.\n")


justinmayer:
Thanks for the fix, Stefan. I will most likely merge https://github.com/getpelican/pelican/pull/1927 soon so it can be included in the upcoming Pelican 3.7 release, at which point you will probably be asked to update this pull request. Apologies for the moving target!  (^_^)

Krastanov:
Thanks for the heads up! Should I understand that the two repos are considered to be tightly coupled and there is no need for backwards compatibility with pelican 3.6.3?

justinmayer:
Backwards compatibility is always preferred but not strictly necessary.

Now that Pelican 3.7 has been released, would you be so kind as to update this pull request to use the new `MARKDOWN` setting?

Krastanov:
Will do in a bit.

justinmayer:
Hi Stefan. Any updates on this?

Krastanov:
@justinmayer I made the change. It might not work on 3.6.4, but it should work on the current version and on versions before 3.6.4. I made the change somewhat blindly (I do not have understanding of what those options do, I just followed the changes described in the pull request that removed MD_EXTENSIONS). I did test it and it seems to work for me.

render_math/math.py
```diff
@@ -269,7 +269,10 @@ def mathjax_for_markdown(pelicanobj, mathjax_script, mathjax_settings):
 
     # Instantiate markdown extension and append it to the current extensions
     try:
-        pelicanobj.settings['MD_EXTENSIONS'].append(PelicanMathJaxExtension(config))
+        if isinstance(pelicanobj.settings.get('MD_EXTENSIONS'), list): # pelican 3.6.3 and earlier
+            pelicanobj.settings['MD_EXTENSIONS'].append(PelicanMathJaxExtension(config))
+        else:
+            pelicanobj.settings['MARKDOWN']['extension_configs'].update({PelicanMathJaxExtension(config):{}})
```
Scheirle: Class instance extensions should be added to the `extensions` list, `extension_configs` is only used for extensions loaded by name. See http://pythonhosted.org/Markdown/reference.html#markdown
```
pelicanobj.settings['MARKDOWN'].setdefault('extensions', []).append(PelicanMathJaxExtension(config)
```

jar1karp:
While you're at it, please fix the process_summary function also. At the moment I'm getting this warning:

WARNING: _get_summary() has been deprecated since 3.6.4. Use the summary decorator instead

and Mathjax is not loaded even if a summary contains math in index pages.

render_math/math.py
Krastanov: Well, that was embarrassing. I will make that change.

Krastanov:
@jar1karp  I will try to track this down.

render_math/math.py
Natim: @Krastanov yes please 👍

Natim:
- better_codeblock_line_numbering/
- extract_toc/
- headerid/
- liquid_tags/
- md_inline_extension/
- render_math/

Maybe we should add a shared helper that plugins can use for that.

jar1karp:
Re: the problem with auto-generated summaries -- it seems that the render_math plugin can't update the summary anymore as the get_summary method is memoized (changed in [this commit](https://github.com/getpelican/pelican/commit/2cb2bf05dfca50afbd3400a344a4aa23f9474dfa)). The plugin gets a truncated summary (i.e. the Mathjax code also gets truncated), does the cleanup and reinserts the Mathjax, but it doesn't actually change the summary. The html will be rendered using the truncated Mathjax so math will not display correctly.

The plugin seems to work correctly if the memoized decorator is commented out for get_summary in contents.py. But I suppose it needs a cleaner fix, maybe a proper mechanism for plugins to be allowed to update summaries?

In addition, if using summary plugin with render_math they have to be loaded in that order to work correctly (render_math would otherwise strip away the special summary marker comments).

render_math/math.py
cranmer: So I tried the patch above and it worked, but the version with 
```pelicanobj.settings['MARKDOWN'].setdefault('extensions', []).append(PelicanMathJaxExtension(config))
 ```
didn't work for me.

**Comment by howthebodyworks** [comment]
I broke the summary issue out into a new issue since it seems to be orthogonal to the basic markdown fix

**Comment by howthebodyworks** [comment]
The alternative pull request referenced above is a minor change - it adds the required math markdown extension instnce  directly to the `extensions`  list rather than adding it to the `extensions_config` list with an empty config. AFAICT from the Markdown docs , this is preferred by Markdown. I believe this change is cosmetic?

**Comment by Krastanov** [comment]
I (hopefully) conclusion:

- this patch should permit the latex plugin to work on <=3.6.3 and >=3.7 (I think the issues that were mentioned with MARKDOWN settings set-up were fixed).
- (does not work on 3.6.4 which uses a dictionary instead of a list?)
- there is a separate issue: summaries do not work because of memoization (see https://github.com/getpelican/pelican-plugins/issues/784)
- somebody took the time to submit a fixed-up patch while I was being lazy (thanks and sorry you had to do it) https://github.com/getpelican/pelican-plugins/pull/831/files

**Comment by justinmayer** [comment]
Thanks to Stefan for the fix and to others for reviewing and assisting. Much obliged! (^_^)

**Comment by szhorvat** [comment]
Is this a separately maintained version of render_math in getpelican/pelican-plugins?  I didn't realize there were several versions.  Here's the same issue in the original repo:

https://github.com/barrysteyn/pelican_plugin-render_math/pull/39

**Comment by jerryasher** [comment]
Regarding @szhorvat's comment, I do wonder, what is the recommended practice when finding/fixing bugs in various Pelican plugins?

I submitted a patch to this bug at the author's repo.
Should I also have submitted a patch (it would have been different as the modules have diverged a bit) over here as well?

(I mean, I guess so, but I didn't want to step on any toes...)

**Comment by drorata** [comment]
When executing `./develop_server.sh start` I still get a warning

```
WARNING: _get_summary() has been deprecated since 3.6.4. Use the summary decorator instead
```

which, as far as I understand, relates to this thread. I fetched all changes to the plugins. The sources in my case can be found [here](https://github.com/drorata/drorata.github.io).  How can I avoid this warning?

PR #791, opened 2016-10-06
Pelican Series is now a PyPI package

cprieto:
Moved the Pelican Series plugin from an inline plugin in this repository to its independent repository to keep a PyPI package for it.

Now the plugin can be installed using PIP: `pip install pelican-series`


kdeldycke:
Indeed: https://pypi.org/project/pelican-series/ ! Now that the plugin has been migrated to its own repository at https://github.com/pelican-plugins/series , I guess we can close this PR.

justinmayer:
Many thanks to @cprieto and @kdeldycke for the assistance here!

PR #798, opened 2016-10-16
liquid_tags > graphviz: Decode base64 output string to UTF-8

solomonhuang:
Fix byte string in python 3

`base64.b64encode` will output byte string in Python 3, and it breaks inline base64 image in HTML.
`decode('utf-8')` will convert a byte string into a string and make  python 2 be happy, too.

```
>>> print(base64.b64encode(b'\01'))
b'AQ=='
>>> print(base64.b64encode(b'\01').decode('utf-8'))
AQ==
```


justinmayer:
Thanks for the fix, Solomon!

PR #800, opened 2016-10-17
Photos: add translation support

Vifespoir:
Hi,
I noticed that the galleries weren't displayed on translated pages. I found that generator.translations has been omitted from `photos.py` > `detect_images_and_galleries()`. Here is a quick patch to serve translation needs on pelican.

By the way thank you so much for such a nice extension!


justinmayer:
Thanks for the enhancement, @Vifespoir.

PR #803, opened 2016-10-20
Add `deadlinks` plugin as a submodule

silentlamb:
This change adds a `deadlink` plugin to manage links to pages that are not available anymore. It allows to replace original link with a link to proper archive.org query. It also allows to modify anchor tags of dead links by adding custom classes or appending bootstrap labels to them.

I hope you find it useful.


justinmayer:
Many thanks for the submission, Marcin.

PR #807, opened 2016-11-16
liquid_tags: Add PyGal generation tag

mitchins:
Generate SVG graphs using PyGal

This has been incorporated into the existing liquid tags.

Syntax is simple as follows (based on the PyGal example graphs).

A pie chart:

```
{%
	pygal {
		"type": "pie",
		"title": "Browser usage in February 2012 (in %)",
		"data" : [
		{"title": "IE",
		 "values": 19.5},
		 {"title": "Firefox",
		 "values": 36.6},
		 {"title": "Chrome",
		 "values": 36.3},
		 {"title": "Safari",
		 "values": 4.5},
		 {"title": "Opera",
		 "values": 2.3}
		]
	}
%}
```

Which renders as follows:

`<div class="pygal" style="text-align: center;"><embed type="image/svg+xml" src="data:image/svg+xml;charset=utf-8;base64,<base64-data>" style="max-width:1000px"></div>
`
It works completely in-line without any external files or resources and is displayed crisply and smoothly.

Attached is a PDF file generated from a .md blog article containing a bunch of graphs.

[Chart Test.pdf](https://github.com/getpelican/pelican-plugins/files/593672/Chart.Test.pdf)

### DIFF
--- a/liquid_tags/pygalcharts.py
+++ b/liquid_tags/pygalcharts.py
@@ -1,0 +1,164 @@
+"""
+pygal Tag
+---------
+This implements a Liquid-style pygal tag for Pelican. JSON is used for the data,
+and you can pass a bunch of pygal's 'config' items through as-is
+
+[1] http://www.pygal.org/
+
+Syntax
+------
+{% pygal
+    {
+        <graph data>
+    }
+%}
+
+Examples
+--------
+{%
+	pygal {
+		"type": "bar",
+		"title": "Test Chart",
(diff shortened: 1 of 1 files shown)


mitchins:
I renamed the file (and hence import tag) to pygalcharts rather than legal because it clashes with legal (being a python library itself) under some circumstances. still use {% pygal ... %} inside the markdown.

an example of the output can now be seen [here](http://blog.mitchellcurrie.com/pelican-pygal-graphs.html)

justinmayer:

PR #808, opened 2016-11-16
Subcategory: Add support for folder-based subcategories.

mfitzp:
This PR adds support for folder-based subcategories, to remove the requirement to edit article files to edit/reorganise categories.

It is implemented based on providing an `PATH_METADATA` parameter to extract `subcategory_path` variable. Setting/unsetting this is sufficient to activate/deactivate folder-based subcategories. Using the regex allows for customisation of folder-subcategory hierarchy — e.g. to exclude specific subfolders, or adjust the hierarchy depth.

I've updated the `README.md` with an example of usage.

### DIFF
--- a/subcategory/README.md
+++ b/subcategory/README.md
@@ -9,6 +9,8 @@ Feeds can be generated for each subcategory, just like categories and tags.
 
 ##Usage##
 
+###Metadata###
+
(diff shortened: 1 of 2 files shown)


justinmayer:

PR #812, opened 2016-12-08
simple_footnotes: Fix unicode errors triggered by latest version of html5lib

johnmarkschofield:
With the current versionof html5lib, the simple_footnotes plugin breaks with
an assertion failure. The root causes is that some strings in simple_footnotes
are unicode and others are str type. For html5lib to be happy, everything has
to be unicode.

This patch fixes the html5lib issue, and introduces no other issues I was able
to find.


justinmayer:

PR #814, opened 2016-12-13
liquid_tags: replace div with span

mothsART:


justinmayer:
Thank you for the contribution, Jérémie.

PR #816, opened 2016-12-13
assets: Make compatible with Pelican 3.7

fbeutel:


justinmayer:
Thanks for the fix, @fbeutel!

PR #817, opened 2016-12-14
liquid_tags: pelican 3.7 compatible

xsteadfastx:
because of the markdown changings.


justinmayer:
Thanks, @xsteadfastx!

PR #818, opened 2016-12-15
Update plugins that use JINJA_ENTENSIONS to the Pelican 3.7 setting

avaris:
Update plugins that use `JINJA_ENTENSIONS` to the Pelican 3.7 setting `JINJA_ENVIRONMENT`

Plugins that were affected:
- assets
- i18n_subsites
- jinja2content


avaris:
BTW, `assets` was somewhat updated, but in a way that made it 3.7-only. I just handled backwards compatibility.

justinmayer:
Many thanks, @avaris. You rock! 🎉

PR #824, opened 2017-01-01
Photos plugin: Add lightbox option to inline photo replacement

wamonite:
* Add `{lightbox}` option for inline images, to insert image compatible with [Lightbox](http://lokeshdhakar.com/projects/lightbox2/), e.g.

        

* For use with other implementations, the gallery and caption attribute names can be set with `PHOTO_LIGHTBOX_GALLERY_ATTR` and `PHOTO_LIGHTBOX_CAPTION_ATTR`.


justinmayer:
Thanks for the contribution, Warren!

Issue #805, opened 2016-10-29
Math Not Rendered For Drafts in render_math Plugin


MarkCWirt:
The `process_rst_and_summaries()` function in `math.py` does not render math if an article is marked as a draft.

The issue is in this code:

```
        for article in generator.articles + generator.translations:
            rst_add_mathjax(article)
            #optionally fix truncated formulae in summaries.
            if process_summary.mathjax_script is not None:
                process_summary(article) 
```

If changed to this:

```
        for article in generator.articles + generator.translations + generator.drafts:
            rst_add_mathjax(article)
            #optionally fix truncated formulae in summaries.
            if process_summary.mathjax_script is not None:
                process_summary(article)
```

it works.

(I'd open a pull request but I note that are a lot of open PRs and this is such a minor change a PR may be overkill.)

howthebodyworks:
Hmm - I also have a fix for this in my branch, 	9686a66
The difference is that my pull request _additinonally_ handles pages but _does not_ handle translations. We shoudl be combining these.
This plugin is getting a little chaotic. @MarkCWirt perhaps you should open a pull request after all?

MarkCWirt:
I can create a pull request, but I saw this one this morning:

https://github.com/getpelican/pelican-plugins/pull/828

So I assume it's not needed?

howthebodyworks:
Yes sorry, I worked up a head of steam to try to tackle a major refactor at #693, and I need this. The pull request should be exactly the same as you suggested @MarkCWirt ;-)

justinmayer:
Presumably fixed by PR 828.

PR #828, opened 2017-01-03
render_math: Handle math markup in drafts

howthebodyworks:
to fix the lack of math markup in drafts mentioned in #805. (CC @MarkCWirt)


justinmayer:
Thank you for the contribution, Dan. (^_^)

PR #831, opened 2017-01-04
render_math: fix for Pelican 3.7 MARKDOWN dict

howthebodyworks:
BEcause I am an idiot I can't persuade github to let me compare with the fork Krastanov:patch-1 as referenced in #787. However, here's my modified patch.


justinmayer:
Perhaps squash the fix-up commit (2859337) into the other logical commit (1e80215)?

justinmayer:
Actually, ignore my last comment. I merged the referenced PR, so if any follow-up work is needed related to this PR, please do so separately.

PR #832, opened 2017-01-04
adding indexing option to representative_image

choldgraf:
This is a quick addition to add the ability to specify an image index to representative image. Instead of just supplying a URL to the article `Image` metadata, you can optionally give an integer. It will assume this integer corresponds to an image in the article. In this way you can include:

```
Image: 3
```
in the metadata, and it would look for the 4th image in the article instead of the 1st. It'll raise a valueerror if you give an index that exceeds the number of images in the article.

Tests were a bit wonky to get working, so somebody should make sure this looks correct.


choldgraf:
closing this bc of no activity

PR #836, opened 2017-01-12
[Bug Fix] Add the mailto:email address in the link

zhangresearch:
Hi! This is trying to fix a bug with email address in the code. In the current version, the Email link only generates the email title and body but not the recipient address.

This change will also require an 'EMAIL_ADDRESS' variable being defined in the pelicanconf.py file.

### DIFF
--- a/share_post/share_post.py
+++ b/share_post/share_post.py
@@ -47,8 +47,9 @@ def share_post(content):
     linkedin_link = 'https://www.linkedin.com/shareArticle?mini=true&url=%s&title=%s&summary=%s&source=%s' % (
         url, title, summary, url
     )
-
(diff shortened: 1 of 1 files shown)


Scheirle:
I think the current behavior is intended, as each visitor want to share this post with different email addresses and not with a fixed email address.

What is the point when all visitors share this post with an email address you specified? 

cc: @talha131 @kernc @muggenhor

muggenhor:
As @Scheirle said: this is definitely not an improvement. You want the e-mail address to be chosen by the user at runtime, not by the administrator at config-time.

You might make it dynamic through javascript or something, but that wouldn't add any (significant) advantage either as this should already be possible to achieve in the mail client.

Maybe you should describe your bug before proposing a fix for it.

PR #844, opened 2017-01-31
bootstrapify: set html parser for BS4.

lll9p:


justinmayer:
Please reach out via the Pelican IRC channel before submitting another PR. Thanks!

PR #848, opened 2017-02-01
gzip_cache: Exclude internally-compressed woff2

janwh:
As per [this gist](https://gist.github.com/sergejmueller/cf6b4f2133bcb3e2f64a#good-to-know), woff2 is compressed also, so it should be excluded from the gzip cache as well.


justinmayer:
Thanks to Jan for the contribution and to Bernard for the review. (^_^)

PR #858, opened 2017-02-10
Math plugin fixed for Pelican 3.7.1

Iv:
_get_summary() is depricated in pelican 3.7

summary property used instead.


jerryasher:
I believe there are reports that due to memoization of the summary, changing `._get_summary()` to `.summary` eliminates the deprecation warning, but leaves a problem where the updated summary as rendered into latex by render-math is not reflected by the property `.summary`

justinmayer:
Many thanks for the contribution, @Kravchenko. Closing in favor of merged PR #973.

PR #859, opened 2017-02-16
sitemap: Add ability to mark individual pages as private

mlissner:
The plugin currently supports using regexes to make a case private. This adds the ability for a page author to mark a page as private, without mucking around in the settings for the entire site.


justinmayer:
Thanks for the enhancement, Mike!

PR #862, opened 2017-02-21
[Pelican Comment System] prep for PyPI release

MinchinWeb:
This is a set of changes to prep this plug in for release to PyPI (to allow installation through `pip`). Specifically:

- add a `setup.py` file
- convert the README and CHANGELOG to ReST, so they render properly on the PyPI site
- move main body of text to `pelican_comment_system` subfolder

To note:

- the version is maintain in `pelican_comment_system/__init__.py`; `setup.py` reads the version from here (via regex) (i.e. the version number is still only maintained in one place)
- a `__init__.py` file is maintained in the root directory. This allows the plugin to continue to be used as it currently is, i.e. from the `pelican-plugins` directory
- the plugin lists Pillow as a dependency in `setup.py`
- there aren't (yet) classifier for Pelican plugins on PyPI

For @Scheirle to review/edit:

- do you want to list an email in `setup.py`?
- I've left it for you to do the actual upload to PyPI


Scheirle:
Great! Thank you very much @MinchinWeb 

I also think this is a good opportunity to split the comment system into its own repository,
so we can do proper releases and so on.

New PR: https://github.com/Scheirle/pelican_comment_system/pull/1

@justinmayer 
This PR can be closed and I will soon add the new repository as submodule or simple link (not sure yet).

justinmayer:
My suggestion is to hold off on both counts. Plan for 4.0 is to move each plugin to separate repository anyway, along with regular PyPI releases. The idea is for repos and PyPI to have shared access, with plugin developers and Pelican developers both having access to ensure continuity and regular updates.

MinchinWeb:
@justinmayer : Do we have a timeline on that? It sounds wonderful, and I would be happy to help with the repo splitting and packaging. I hesitate to hold on this pull request though without a short timeline because moving the plugins onto PyPI has been kicking around for [at least 4 years now](https://github.com/getpelican/pelican-plugins/issues/39)...

As an interim measure, is there a way to split this off the main repo in a way that makes future efforts easier?

PR #867, opened 2017-03-04
series: Fix bug in sorting with numbers > 9

ix5:
The `operator.itemgetter()` compare function reads the `series_index` metadata as a string instead of a number. This leads to e.g. 20 being sorted before 3, because it begins with a 2.

This commit converts the number to an `int` beforehand.

The `key=lambda` syntax works even with older python versions, tested on 2.7 and 3.6.


ix5:
Closed in favor of upstream changes

PR #872, opened 2017-03-19
Video Privacy Enhancer: Add Python 3 support

publicus:
I upgraded my Pelican virtualenv this weekend to Python 3 and Pelican 3.7.1, and realized that the Video Privacy Enhancer breaks with the upgrade. Thus, I added use of the `six` package to support both Python 2 and 3. I then tested re-generating my site with Python 2.7.12 (with Pelican 3.6.3) and Python 3.4.5 (with Pelican 3.7.1).

### DIFF
--- a/video_privacy_enhancer/video_privacy_enhancer.py
+++ b/video_privacy_enhancer/video_privacy_enhancer.py
@@ -22,9 +22,11 @@ import re # For using regular expressions.
 import logging
 logger = logging.getLogger(__name__) # For using logger.debug() to log errors or other notes.
 
-import urllib 
+import six
+# import urllib.request
+import six.moves.urllib.request 
 
-import video_service_thumbnail_url_generating_functions as video_thumbnail_functions # These are functions defined in 'video_service_thumbnail_url_generating_functions.py', which is located in the same directory as this file. 
+from . import video_service_thumbnail_url_generating_functions as video_thumbnail_functions # These are functions defined in 'video_service_thumbnail_url_generating_functions.py', which is located in the same directory as this file. 
 
 """
 END OF LIBRARIES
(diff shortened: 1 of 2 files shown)


publicus:
Actually, I just realized that I didn't include one final commit when I squashed to open this PR. I'll update the PR in just a bit... Please hold on this until then.

publicus:
Ok, the additional commits have now been added. I would be grateful for your consideration of this Pull Request. Thank you for your work!

video_privacy_enhancer/video_privacy_enhancer.py
```diff
@@ -115,38 +117,30 @@ def download_thumbnail(video_id_from_shortcode, video_thumbnail_url, video_servi
...
-		logger.debug(video_service_name)
-		
 		logger.debug("Video Privacy Enhancer Plugin: The name of the current service being processed is '" + video_service_name + "'")
 		
 		shortcode_to_search_for_not_including_exclamation_point = video_service_information['shortcode_not_including_exclamation_point']
 		logger.debug("Video Privacy Enhancer Plugin: Currently looking for the shortcode '" + shortcode_to_search_for_not_including_exclamation_point + "'")
 		
 		function_for_generating_thumbnail_url = video_service_information['function_for_generating_thumbnail_url']
-		logger.debug("Video Privacy Enhancer Plugin: Using the following function name to get thumbnails for the current video service:")
-		logger.debug(function_for_generating_thumbnail_url)
```
justinmayer: Is there some reason you removed this debugging information?

publicus: I realized that it was throwing an error in the debug output. When I originally wrote it a couple of years ago, I thought that it would output the name of the function being used; as I was putting this PR together, though, I realized that it doesn't. Thus, I removed it to clean things up while I was thinking about it.

justinmayer: Got it. Thanks for the explanation!

justinmayer:
Thanks for the enhancement, Jacob!

video_privacy_enhancer/video_privacy_enhancer.py
publicus: Thank you for your work on the project! : )

publicus:
Thank you!

PR #873, opened 2017-03-19
Add support for Tipue Search plugin for current Pelican version.

publicus:
I'm not sure exactly in which version, but somewhere around Pelican 3.6.2, I noticed that the Tipue Search plugin was generating search results, but with URLs of value "Undefined". I noticed that adding a `loc` attribute got the plugin working again. I think that the `url` attribute is unnecessary now, but am not sure, and so have left it in, as well.


jorgesumle:
I'm using the same code from this repo of Tipue search and Pelican 3.6.3 and don't have this problem in my own website: http://www.freakspot.net/, or at least I haven't found it out yet. Could you explain why and when does this happen better?

publicus:
I'm not sure yet whether this is an issue just related to certain themes. It's come up in conversation in two locations that I've found, [here](https://github.com/talha131/pelican-elegant/issues/147) and [here](http://blog.siphos.be/2015/08/updates-on-my-pelican-adventure/). With the mnmlist theme, it broke for me in Pelican 3.6.2 or 3.6.3. Since defining both `url` and `loc` isn't a major change, and, to my understanding, shouldn't break compatibility with sites that currently work, it seems a useful addition to me.

archeranimesh:
What is the status on this pull request, It is fails for me while using the elegant theme, and this plugin, the search is working, but the URL are all undefined.
[my Site.](http://archerimagine.com/)

publicus:
It sounds to me as though in some themes, `url` is used, and in some, `loc` is used. @jorgesumle, would you be willing to confirm that in your setup, where `url` is used, adding `loc` (without removing `url`) doesn't break your setup? If both can co-exist (as has been the case in my experience -- I have both included in my setup with Pelican 3.7.1), I think that that would be good evidence for merging this Pull Request.

iranzo:
Related https://github.com/Pelican-Elegant/elegant/issues/147

iranzo:
@justinmayer hey, can we have this reviewed and merged?

We do have alternate way in our plugin but as it was said above, each theme is handling it in a different way so this patch will address both approaches.

Thanks

publicus:
@justinmayer, thank you for all of the work that you do for this project!

PR #877, opened 2017-03-31
Just Table plugin: Update submodule

burakkose:
Use latest version of just table plugin and update the plugin description


burakkose:
thanks @justinmayer

PR #882, opened 2017-04-10
Update MathJax CDN, closes #881

### LINKED ISSUE #881: [notebook] MathJax CDN changed again
Announcement here: https://www.mathjax.org/cdn-shutting-down/

There is a new CDN taking over but this line needs to be updated: https://github.com/getpelican/pelican-plugins/blob/master/liquid_tags/notebook.py#L170

isms:


justinmayer:
Thanks for the fix, Isaac!

PR #884, opened 2017-04-15
Update render_math plugin to use cdnjs

### LINKED ISSUE #881: [notebook] MathJax CDN changed again

mirzal:
As MathJax is closing its CDN, users are advised to use local version or another CDN, with cdnjs being preferred at the moment. Update link in plugin configuration and really close #881


justinmayer:
Thanks for the fix, @mirzal. Would you please change the scheme to an explicit `https://`? In 2017, the current protocol-relative method is widely considered to be an anti-pattern.  :smile:

mirzal:
@justinmayer Sure, no problem. Done in c11f45b.

justinmayer:
Excellent. Thank you!

PR #891, opened 2017-05-08
Photos: Handle photos without EXIF data more gracefully

AntoineVe:
If we try to use EXIF manipulation, and a picture as no EXIF data, thumbnails and resized pictures are not created.
With this little error handling, a debug info is showed and everything continue, creating resized and thumbnails pictures.


justinmayer:
Thanks for the contribution, Antoine! 😄

PR #892, opened 2017-05-08
Add pelican-plotly plugin

ran404:
Add pelican-plotly plugin.

This plugin allows embedding of offline plotly charts (json files) using the [plotly] tag.

See [this blog](http://www.arraystream.com/blog/deep-selection.html) for an example of the output.

Thanks

Ran


datavistics:
This addition would be much appreciated!

bittner:
The link of the original description is dangling. Can you fix that and rebase the PR?

ran404:
fixed and rebased PR, could you pls check?

PR #899, opened 2017-05-24
Add github wiki plugin and update readme

rhyst:
A plugin that converts a flat GitHub wiki into a structured read only wiki on your site.


justinmayer:
Apologies for the delay in merging. Many thanks for the contribution, @rhyst!

PR #900, opened 2017-05-28
Add author_images plugin

WPettersson:
This plugin adds support for author.image and author.avatar properties, based on files found with appropriate naming. It searches the AUTHOR_IMAGES (or _AVATARS) path for files named appropriately, and then adds the locations of these files (if found) to author.images (or author.avatars) for use when displaying articles.

I'm not convinced it's ready for merging, I have a few questions which I cannot answer myself.
I don't know if I'm calling my function with the right signal. I've dug through the Pelican code, and tested many options, and I just can't work out how it all glues together. Currently it uses article_generator_finalized, but this feels too late as I not only have to iterate over all authors in the generator, but also all authors in each article in the generator.

I'm also finding file names of images/avatars by running a sha256sum over the author name, and then appending various extensions. Any hash should do, even MD5 should be fine as the worst that should happen is the wrong image is displayed, but I didn't see any significant downside in using a more recent hashing algorithm. I'm happy to be told to use something else here. And I decided not to just use the author name directly as I imagine some webhosts might not allow full unicode filenames, and this also avoids any issues with spaces in filenames.

### DIFF
--- a/author_images/README.md
+++ b/author_images/README.md
@@ -1,0 +1,55 @@
+# author_images
+
+This Pelican plugin adds support for author images and avatars. You may choose
+to display one or the other.
+
+## Configuration
+
+Add the directory to the base plugins directory to `PLUGIN_PATHS` in
+`pelicanconf.py`, and then add `author_images` to the `PLUGINS` list. For example,
+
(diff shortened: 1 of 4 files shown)


justinmayer:
Many thanks for the contribution, William!

PR #903, opened 2017-06-01
Move bootstrap-rst from submodule to in-tree

aqw:
I no longer have intent to further this plugin, so I think it's best to move it in-tree where it's more easily accessible for others.

Since everything is squashed, the changes from the last version are:
  * use relative imports in bootstrap.py (thanks to @davidchall)
  * updated maintainer email
  * very small documentation fixes

If you have any questions, let me know. :-)

---Alex


justinmayer:
Many thanks, Alex!

PR #906, opened 2017-06-08
Modify os module import syntax

gavinzbq:
Under the original circumstance, test method `path` can be confused with `path` attribute of `os` module. 

By altering import syntax this way, it is immediately obvious `path` attribute is from `os`.


justinmayer:
Hi @gavinzbq. Thank you for your fix. Would you please remove the spurious `.swp` file and fix-up into a single commit?

(I have added entries to the project `.gitignore` file in order to mitigate this occurrence in the future: b3b8d36)

gavinzbq:
Thanks @justinmayer, I have created a new pull request which only fixes `thumbnailer` test code.

justinmayer:
In the future, please re-use existing pull requests instead of creating new ones. See: https://github.com/getpelican/pelican/wiki/Git-Tips

PR #907, opened 2017-06-09
Thumbnailer: Modify os module import syntax

gavinzbq:
Under the original circumstance, test method `path` can be confused with `path` attribute of `os` module.

By altering import syntax this way, it is immediately obvious path attribute is from os.


justinmayer:
Please squash into a single commit, as mentioned here: https://github.com/getpelican/pelican/wiki/Git-Tips

gavinzbq:
I have done squashing commits. Please let me know if there is any problem @justinmayer . Thanks :)

justinmayer:
Thanks again, @gavinzbq.

PR #910, opened 2017-06-26
[render_math] Fix NameError in Python 3

diazona:
This PR is to fix the `NameError` I get when using MathJax extensions. The error comes from the line

    filter(lambda string: isinstance(string, basestring))

Because `filter()` is a generator in Python 3, the lambda function doesn't actually called at the point of the `filter()` statement, it gets called later, outside the enclosing `try` block. I've changed this to identify the relevant string type in advance, at module level, so the new call to `filter()` doesn't raise this error.

I haven't created an issue in advance because this seemed like a very straightforward and simple fix.

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -45,6 +45,12 @@ try:
 except ImportError as e:
     PelicanMathJaxExtension = None
 
+try:
+    string_type = basestring
+except NameError:
+    string_type = str
+
+
 def process_settings(pelicanobj):
     """Sets user specified MathJax settings (see README for more details)"""
 
@@ -90,10 +96,7 @@ def process_settings(pelicanobj):
         # and 3 of python
 
         if key == 'align':
-            try:
-                typeVal = isinstance(value, basestring)
-            except NameError:
-                typeVal = isinstance(value, str)
+            typeVal = isinstance(value, string_type)
 
             if not typeVal:
                 continue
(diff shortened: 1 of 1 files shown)


avaris:
FWIW, [`six.string_types`](https://pythonhosted.org/six/#six.string_types) is also an option.

justinmayer:
@avaris: Any preference as to whether to merge this PR or #927?

diazona:
@avaris Ah, right, I didn't think of `six.string_types`. If you'd rather do it with `six` I can make the change in another commit, if that's necessary. (I'm not sure how easy it is for other people to add commits)

justinmayer:
@diazona: Probably easier for PR author to make changes, methinks.

avaris:
@justinmayer not really. They are pretty much identical. This one has the merit of being first :).

@diazona It's not a *must*. If you want to do the work of changing it and rebasing to a single commit, then sure. But, it's a minor improvement. I'm OK with the way it is now.

justinmayer:
Merging for now. Follow-up pull requests welcome, whether that be by @diazona, @barrysteyn, @oulenz, @pwaqo, or @avaris. Thanks, all!

diazona:
@avaris No trouble at all. I put it on a new branch `fix_string_error` in my fork - see diazona@62b79d085b3c04008a99c6a01926e7895f76c67c. Up to you and @justinmayer whether you want to use this one instead.

PR #914, opened 2017-07-07
Jinja2htmlcontent

radupotop:
This plugin enables the use of Jinja2 directives inside Pelican 
**HTML** articles and pages. 

This plugin is a shameless rip-off the `jinja2content` plugin.
It was adapted in order to work with `.html` content files.
This enables creating complex static sites with Pelican using html
instead of Markdown, and having components split into macros and fragments.

### DIFF
--- a/jinja2htmlcontent/README.md
+++ b/jinja2htmlcontent/README.md
@@ -1,0 +1,67 @@
+# Jinja2 in HTML articles and pages
+
+This plugin enables the use of Jinja2 directives inside Pelican 
+**HTML** articles and pages. 
+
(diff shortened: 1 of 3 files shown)


aqw:
@wooptoo How does this compare with the jinja2content plugin?

kdeldycke:
This plugin duplicates the feature set already implemented by https://github.com/pelican-plugins/jinja2content . As the latter has already been migrated out of the mono-repo and packaged to PyPi, I propose to close this PR.

justinmayer:
Thanks to @radupotop and @kdeldycke for your work here. ✨

PR #917, opened 2017-07-23
Add RDF plugin to parse vocabularies and process them as a new media type

NSeydoux:
This plugin offers a flexible mechanism to write lightweight documentation about rdf vocabularies, based on sparql queries. The core idea is to allow the user to both access right away to essential informations about the vocabulary, but also to allow him to extract custom information about its vocabulary. The usage of this plugin is demonstrated for instance on the page https://irit_ut2j.frama.io/, which is still a work in progress but shows the integration of the plugin into a gitlabl pages process.


justinmayer:
Thanks for the contribution, @NSeydoux!

PR #919, opened 2017-08-05
Add ERT (estimated reading time) plugin

nogaems:
A plugin for estimating time that could be spent for an article reading.


justinmayer:
Thanks for the contribution, @nogaems.  😄

PR #921, opened 2017-08-07
Latex macros added

DjalelBBZ:
Most of the code was imported from https://github.com/barrysteyn/pelican_plugin-render_math/


justinmayer:
Many thanks for the contribution, @DjalelBBZ. What is the benefit to calling a CDN with anything other than HTTPS? Why even have HTTP as an option? I don't see the benefit.

kdeldycke:
FTR, `render_math` plugin has been moved to its own repository at: https://github.com/pelican-plugins/render-math

PR #925, opened 2017-08-13
Interlinks: patch to fix #920

### LINKED ISSUE #920: [Interlinks] Last update to interlinks has broken <img> tags
Unfortunately, the accepted PR request described here breaks  tags in the Interlinks plugin: https://github.com/getpelican/pelican-plugins/commit/e4c2f1691ae243474067c3844339d82c44657c56

The error appears to be on line 56, where `str(link)` refers to a variable that doesn't exist.  

To reproduce this:

 1. Install latest Pelican 3.7 and configure Interlinks plugin (below)
 2. Add the following to a markdown document in content: `text`
 3. `make html` and observe that there is one execution error _per link_

# Interlinks Configuration #

Add the following to pelicanroot:

    INTERLINKS = {
       'imgroot': '/postimages/',
    }

aquinzi:
Sorry for the very looong wait to fix this silly bug.

### DIFF
--- a/interlinks/interlinks.py
+++ b/interlinks/interlinks.py
@@ -53,7 +53,7 @@ def parse_links(instance):
 				if name in interlinks:
 					hi = url.replace(name+">",interlinks[name])
 					img['src'] = hi
-				content = content.replace(old_tag, str(link))
+				content = content.replace(old_tag, str(img))
 
 		instance._content = content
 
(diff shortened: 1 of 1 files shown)


justinmayer:
Thanks for the fix, @aquinzi.

@andyli / @mlabbe: Hopefully this will address the problems you reported. (^_^)

PR #927, opened 2017-08-20
[render_math] fix python3 error with basestring where statement in try block wasn't evaluated until later

oulenz:
This fixes a bug in lines 164--167 where in Python 3, `isinstance(string, basestring)` doesn't trigger the exception because the statement isn't evaluated until later.

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -45,6 +45,11 @@ try:
 except ImportError as e:
     PelicanMathJaxExtension = None
 
+try:
+    STRING_CLASS = basestring
+except NameError:
+    STRING_CLASS = str
+
 def process_settings(pelicanobj):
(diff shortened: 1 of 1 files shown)


oulenz:
This pertains to the render_math plugin, and @barrysteyn is probably best positioned to review this.

pwaqo:
I think this is a duplicate of #910

oulenz:
Hm, yes. Happy to see that we came upon the same solution to fix this. But it doesn't inspire me with confidence that that pull request is still open, it's a straightforward bug fix. @justinmayer @barrysteyn

barrysteyn:
Sorry, I will look at it today

justinmayer:
Closed in favor of prior PR #910. Many thanks to @oulenz for trying to get this fixed! (^_^)

PR #931, opened 2017-08-26
Thumbnailer: add optional input file name regex setting

tomka:
With the help of the THUMBNAIL_INCLUDE_REGEX setting it is now possible to restrict thumbnailing to a files matching this regular expression. The setting is expected to be a string. If available, a regular expression is created. The default behavior is still that all files not starting with a dot are respected.

For instance, this can be set to only match files ending in "jpg", "jpeg" or "png" one can use:

THUMBNAIL_INCLUDE_REGEX = r'(jp[e]g|png)$'

This is useful to exclude other files that sometimes go along with images, e.g. titles.


justinmayer:
Many thanks for the contribution, Tom!

PR #940, opened 2017-10-06
AsciiDoc: Support both legacy and asciidoctor via CLI subprocess

jeffrimko:
Updated the `asciidoc_reader` to use either the original `asciidoc` CLI utility or the new `asciidoctor` utility. This change no longer uses the legacy API but instead directly calls the CLI utility to render text. The main benefit of this change is that Python 2 is no longer a requirement and this plugin will support both Python 2 and 3.


avaris:
Why `auxly` instead of `subprocess`, especially when it [comes](https://pypi.python.org/pypi/auxly/0.3.4) with a warning?
>The status of this project is pre-alpha. This project is not yet suitable for use other than testing.

jeffrimko:
Good point about the warning. It was used it to get the functionality working quickly and I did have some second thoughts about using a third party library in a plugin. If you'd prefer, I can refactor to only use `subprocess`.

Full disclosure: I am the developer of Auxly. The warning is mostly because I have not yet had a chance to thoroughly test potentially dangerous functions like `delete()`.

justinmayer:
Thanks for the contribution, Jeff. I agree with @Avaris; it's usually best to avoid adding external dependencies unless absolutely necessary, and `subprocess` should suffice for the use case at hand. Would you please post a comment here when the refactor is ready for review?

jeffrimko:
Updated the plugin to remove third party dependencies. Changed the logic around a bit to write the contents of the `pelican_open(source_path)` to a temporary file which seems a bit hacky but was necessary to avoid issues with the relative `source_path`. Maybe there is a better way to handle this. I took a quick look through `pelican.utils` but didn't see anything that would return an absolute `source_path`. I'll take a closer look later but for now this plugin should be functional.

jeffrimko:
Upon further review, no longer seeing the strange relative path behavior I thought I saw previously. In fact, looking through the Pelican code suggests that the `source_path` is always absolute. Tested this plugin update with Python 2.7 and 3.6 on Windows without issue.

justinmayer:
@avaris: Any further comments on this pull request?

@jeffrimko: Might make sense to [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits)?

jeffrimko:
Okay, I think the rebase/squash is all ready.

haridas:
+1 for this addition. Looking forward to see it in plugins list.

nkalvi:
@justinmayer Could you please merge? I'd like to test the changes. Thanks.

justinmayer:
Many thanks for your work on this, Jeff.

nkalvi:
@justinmayer @jeffrimko Thank you!

avaris:
FYI, you can [checkout PRs](https://help.github.com/articles/checking-out-pull-requests-locally/). You don't need to wait for a merge to test it.

nkalvi:
@avaris Thanks - didn't know it. Will do in the future.

PR #945, opened 2017-10-15
[liquid-tags] specify in conf which params should be passed sub-plugins

epaulson:
Starting in #321, liquid tag sub-plugins (e.g. a new tag) that needed a config entry had to modify two arrays in the liquid-tag/ directory. This is sort of a pain for a sub-plugin that's not in the tree.

Here's a quick fix that looks into a user's pelicanconf.py for a LIQUID_CONFIGS array and makes those config files entries available to the plugin as well, ala:

```python
LIQUID_CONFIGS = (('PATH', '.', "The default path"), ('SITENAME', 'Default Sitename', 'The name of the site'), ('FOO', 'Default for foo', 'A nonsense test'))
```
(it's parameter, a default value, and then help text)

At some point, perhaps the hard-coded settings in LT_CONFIG could be moved into a default config file setting, but this PR should happily coexist with the settings that are already in LT_CONFIG out of the box, and any changes to LT_CONFIG that a local installation may have done. 

My use case is a new tag that fetches URLs and extracts OpenGraph info before rendering the link (what [slack calls 'unfurling' a message](https://api.slack.com/docs/message-link-unfurling) or how posting a link on FB shows a preview) - I don't want to do it for every link in all posts, so I have a liquid tag for the links I do want to unfurl. I want to build a cache of OG info so I don't refetch every link any time I rebuild my site, and I need to know where to store the cache, so I need 'PATH'. This PR gives me and other sub-plugin authors a way to get new params into the sub-plugin with less fuss.

Thanks!

### DIFF
--- a/liquid_tags/Readme.md
+++ b/liquid_tags/Readme.md
@@ -220,6 +220,17 @@ comment line `# <!-- collapse=False -->` will be expanded on load but
 can be collapsed by tapping on their header. Cells without collapsed
 comments are rendered as standard code input cells.
 
+## Configuration settings in custom tags
+
+Tags do not have access to the full Pelicans settings, and instead arrange for 
+the variables to be passed to the tag.  For tag authors who plan to add their 
+tag as in-tree tags, they can just add the variables they need to an array in 
+`mdx_liquid_tags.py`, but out-of-tree tags can specify which variables they 
+need by including a tuple of (variable, default value, helptext) in the 
+user's `pelicanconf.py` settings:
+
+    LIQUID_CONFIGS = (('PATH', '.', "The default path"), ('SITENAME', 'Default Sitename', 'The name of the site'))
+
 ## Testing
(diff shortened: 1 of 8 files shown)


justinmayer:
Thanks for the contribution, Erik. Would you consider adding docs/tests?

epaulson:
@justinmayer I have updated the README and included a test. 

I disabled the test_notebook test - it looks like that has been broken since a refactoring in the main pelican tree changed how the pelican code was installed, which left out the test code and supporting modules, which meant that tox didn't get everything needed into the test tree:

https://github.com/getpelican/pelican/commit/bd5dfa0b3ec9560a1a453907f9f3767b9b89ea01#diff-2eeaed663bd0d25b7e608891384b7298

justinmayer:
Many thanks for the improvements, Erik!

PR #949, opened 2017-10-19
New Org reader

Postroutine:
In this pull request:
- Rename the actual org_reader to org_emacs_reader, because this plugin need emacs.
- Write a new org_reader that don't need external sowftware. It use orgco: https://github.com/paetzke/orgco


Postroutine:
@justinmayer: Any news?

jappeace:
Oh wow, I was thinking of doing this myself but you already did this!
You should probably advertise why this is necessary and I suspect it has something to do with speed. (I found the 'old' emacs based one to be unnacaptably slow).
I think overwriting the old org_reader may not be a good idea as other people could use emacs specific dependencies, then when they upgrade they'll find the plugin doesn't work as expected anymore.

PR #955, opened 2017-11-03
Added support for attached images

mitchins:
Current implementation did not work with {attach} ( or even images without an absolute path), because:
1. Assumed the item existed at output already
2. Could not deal with {attach} in name, did not remove and threw warning

The solution:

Does not assume file path is absolute, and does not require image to be copied yet.

If detects {attach} in filename-only then it will look relative to the current data file (.md for example) for reading properties.

 **Otherwise maintains the way it works now, have tested with data for example and the conditional only triggers when {attach} is seen, all else is as per existing.**

### DIFF
--- a/better_figures_and_images/better_figures_and_images.py
+++ b/better_figures_and_images/better_figures_and_images.py
@@ -43,39 +43,46 @@ def content_object_init(instance):
             logger.debug('Better Fig. img_path: %s', img_path)
             logger.debug('Better Fig. img_fname: %s', img_filename)
 
-            # Strip off {filename}, |filename| or /static
-            if img_path.startswith(('{filename}', '|filename|')):
-                img_path = img_path[10:]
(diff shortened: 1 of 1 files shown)


mitchins:
@justinmayer I think this goes to you normally.

justinmayer:
Hi Mitchell. Is there a reason you closed this pull request? Also, it seems the source repository has disappeared.

mitchins:
Thanks Justin,
There should be a newer one.
I didn’t realise I originally created it on a pull request I made 6 months ago hence it was out of date! Do you see the new one? (Should be #956)

Better safe than sorry, so I checked out a fresh copy, as I’m not sure how to rebase on GitHub with forks)

mitchins:
https://github.com/getpelican/pelican-plugins/pull/956

PR #956, opened 2017-11-03
better_figures_and_images: Add support for {attach} images

mitchins:
(Earlier request was against an older fork)

Current implementation did not work with {attach} ( or even images without an absolute path), because:

Assumed the item existed at output already
Could not deal with {attach} in name, did not remove and threw warning
The solution:

Does not assume file path is absolute, and does not require image to be copied yet.

If detects {attach} in filename-only then it will look relative to the current data file (.md for example) for reading properties.

Otherwise maintains the way it works now, have tested with data for example and the conditional only triggers when {attach} is seen, all else is as per existing.

@justinmayer please let me know your thoughts.


justinmayer:
Thank you for the contribution, Mitchell.  (^_^)

PR #958, opened 2017-11-07
assets: ASSET_SOURCE_PATHS should be list instead of tuple

ssteinerx:
Tuple causes error when used, must be list.


justinmayer:
Thank you for the contribution, Steve.  (^_^)

PR #959, opened 2017-11-11
txt2tags_reader: Add --no-headers option

fgallaire:
Important improvement of the txt2tags html output (remove not needed header and footer).


justinmayer:
Thank you for the contribution, Florent.  (^_^)

PR #960, opened 2017-11-12
filetime_from_git: Handle non-ASCII paths

### LINKED ISSUE #328: filetime_from_git gives encoding errors
The `filetime_from_git` plugin together with paths with unicode characters gives me errors like the following:

```
WARNING: Could not process inventar/køleskab.rst
'ascii' codec can't encode character u'\xf8' in position 58: ordinal not in range(128)
```

Changing [line 32](https://github.com/getpelican/pelican-plugins/blob/master/filetime_from_git/filetime_from_git.py#L32) to

```
path = content.source_path.encode('utf-8')
```

solved the issue for me on Python 2.7.

mbudde:
Paths with non-ascii characters would give encoding errors.

Fixes #328

### DIFF
--- a/filetime_from_git/git_wrapper.py
+++ b/filetime_from_git/git_wrapper.py
@@ -39,7 +39,7 @@ class _GitWrapperCommon(object):
         :returns: True if path is managed by git
         '''
         status, _stdout, _stderr = self.git.execute(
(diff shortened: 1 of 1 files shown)


justinmayer:
Many thanks, Michael!

PR #961, opened 2017-11-15
Update i18n_subsites readme

mandaris:
The plugin calls for the extension for the command “gettext” and
doesn’t write the text without it.


justinmayer:
Thanks, @mandaris!

PR #963, opened 2017-11-19
Allow photos plugin thumbnails to be cropped as squares

NicolasLM:
Arranging thumbnails of pictures with different aspect ratios in page is always a problem. This merge request allows to use the solution picked by Instagram: cropping thumbnails as squares.

An example of this feature can be found on my [personal page](https://lemanchet.fr/photography.html).


voidplayer:
This is a great idea. I hope it gets merged soon

Ive been having this same problem, but I need cropped rectangles instead of squares

Could it be possible to specify one dimension and crop the other one?

justinmayer:
Thanks for the contribution, Nicholas.

Issue #928, opened 2017-08-22
Mathjax broken: No MathJax URL specified


walchko:
When I try to build my site, it crashes because of render_math. This use to work nicely.

```bash
kevin@Logan pelican $ make html
pelican /Users/kevin/github/pelican/content -o /Users/kevin/github/pelican/www -s /Users/kevin/github/pelican/pelicanconf.py -D
DEBUG: Pelican version: 3.7.1
DEBUG: Python version: 3.6.1
DEBUG: Adding current directory to system path
DEBUG: Temporarily adding PLUGIN_PATHS to system path
DEBUG: Loading plugin `render_math`
DEBUG: Registering plugin `render_math`
DEBUG: Loading plugin `code_include`
DEBUG: Registering plugin `code_include`
DEBUG: Loading plugin `sitemap`
DEBUG: Registering plugin `sitemap`
DEBUG: Restoring system path
DEBUG: Found generator: <class 'sitemap.sitemap.SitemapGenerator'>
DEBUG: Template list: ['!simple/archives.html', '!simple/article.html', '!simple/author.html', '!simple/authors.html', '!simple/base.html', '!simple/categories.html', '!simple/category.html', '!simple/gosquared.html', '!simple/index.html', '!simple/page.html', '!simple/pagination.html', '!simple/period_archives.html', '!simple/tag.html', '!simple/tags.html', '!simple/translations.html', '.DS_Store', '_includes/.DS_Store', '_includes/footer.html', '_includes/header.html', '_includes/pages.html', '_includes/pagination.html', '_includes/social.html', '_includes/tags.html', 'archives.html', 'article.html', 'author.html', 'authors.html', 'base.html', 'categories.html', 'category.html', 'gosquared.html', 'index.html', 'page.html', 'pagination.html', 'period_archives.html', 'tag.html', 'tags.html', 'translations.html']
DEBUG: Read file blog/arch_linux/arch.rst -> Article
DEBUG: Signal article_generator_preread.send(ArticlesGenerator)
DEBUG: Signal article_generator_context.send(ArticlesGenerator, <metadata>)
DEBUG: Read file blog/computers/benchmark.rst -> Article
DEBUG: Read file blog/computers/clamav.rst -> Article
DEBUG: Read file blog/computers/curl.rst -> Article
DEBUG: Read file blog/computers/ethernet.rst -> Article
DEBUG: Read file blog/computers/ffmpeg.rst -> Article
[...]

PR #966, opened 2017-11-21
render_math/math.py: fix build for pelican

senya:
fix https://github.com/getpelican/pelican-plugins/issues/928
Fix is provided by jcbagneris.


MicroJoe:
Any update on this? My build is broken with recent Pelican install.

PR #967, opened 2017-11-21
Add jpeg_reader plugin to create articles from JPEG metadata

mitchins:
TL;DR, it's essentially a gallery or image page generator based on the JPEG metadata. Name files '.jpeg_article' so that article static content doesn't get dragged in unintentionally.

See README.md for more detail. Has basic pytest cases. Creates full size image and thumbnail at the destination. Requires Exiv2 to be installed, doesn’t use the bindings which will not install on python3 for me, uses process output.

It's rudimentary, but meets all the criteria of metadata required for pelican (and then some such as tags, template, category). Ideally would be better with a python library that can read IPTC natively, but there are few of these (fewer if any that are still supported/maintained), and EXIV2 bindings are currently broken during install. There seems to be a certain length when EXIV2 binary just says "Binary values suppressed" during output for the description field, but there's little we can do. PIL library allows EXIF reading via the _get_exif() and so potentially I might use the for EXIF, and read only IPTC from EXIV2, only downside is referencing private methods violates 'style'.

Also made further tweaks to the problematic 'better figures and images' library: (i) If there's already styles there, don't do the output routine as this would naturally add multiple attributes in the style field and is thus undefined. (ii) If the image doesn't exist in the source location, check the output location.


mitchins:
@justinmayer I think it's been overlooked as there is no reviewer. I'd be keen to get some feedback, I successfully used this to generate a photo gallery in pelican and would be keen to see it merged, or improve the code/style if thats what you want.

justinmayer:
Many thanks for your patience and for the contribution, Mitchell.

PR #969, opened 2017-11-29
Fix sub-category names construction

suriya:


justinmayer:
Thanks for the contribution. Would be helpful if you included a description of the problem that this change intends to solve. Would you mind elaborating?

suriya:
The original code does not work. The patch fixes it.

The patch shows that the original code has a call to `subcategory.strip()` but never uses the stripped value. The call to `strip` is essentially dead code. The patch corrects this by using the stripped value.

justinmayer:
Thanks for the clarification and the fix, Suriya.

PR #970, opened 2017-11-29
Fix builtin commands handling in the linker plugin

suriya:


justinmayer:
Thanks for the contribution. Would be helpful if you included a description of the problem that this change intends to solve. Would you mind elaborating?

PR #850, opened 2017-02-04
fixed warning _get_summary() deprecated in render math

bassdeveloper:
Render Math plugin produced the following warning : 
```shell
# WARNING: _get_summary() has been deprecated since 3.6.4. Use the summary decorat
# or instead
```

Fixed it by simply changing the function to get_summary() instead of deprecated function _get_summary().

PR #973, opened 2017-12-30
Render Math: Workaround for processing summaries

lhofmann:
Since Pelican 3.7, render_math no longer is able to process summaries.

Simply fixing the deprecation warning (#850, #858) does not solve this issue, as the `get_summary` function is memoized using `pelican.util.memoize`.
As a workaround, the memoize instance is acquired by inspecting the result of `memoize.__call__` and  its cache subsequently is cleared.

The following assumptions are made, which might not hold for future versions of pelican:

- `article.get_summary` is an instance of `functools.partial`
- the function bound by this instance is a bound method of an `pelican.util.memoize` instance
- `pelican.util.memoize` has a member `cache` that stores the memoization

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -187,7 +187,7 @@ def process_summary(article):
     """Ensures summaries are not cut off. Also inserts
     mathjax script so that math will be rendered"""
 
-    summary = article._get_summary()
+    summary = article.summary
     summary_parsed = BeautifulSoup(summary, 'html.parser')
     math = summary_parsed.find_all(class_='math')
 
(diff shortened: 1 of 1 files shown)


lhofmann:
Could you please provide the exact steps to reproduce this?
I'm not getting any errors and the indentation looks fine to me.

sus007:
After a clean install/configuration of **`pelican`** and inclusion of **`render_math`** plugin in my **`pelicanconf.py`**, I got the following error.
> WARNING: _get_summary() has been deprecated since 3.6.4. Use the summary decorator instead  
Then after applying your solution i.e after replacing **`._get_summary()`** with just **`.summary`**, I started getting **`indentation error`** for that specific altered line.  
#### I'm a Python noob, so I'm sorry if I sound foolish. But wasn't that your workaround fix for the issue. Is there something else that I have to change in my **`pelicanconf.py`** or **`math.py`** ?

lhofmann:
Please make sure to apply the entire [patch](https://github.com/getpelican/pelican-plugins/pull/973/files). You can find the full modified `math.py` file [here](https://github.com/lhofmann/pelican-plugins/blob/render_math-summary-workaround/render_math/math.py).

justinmayer:
Many thanks for the fix @lhofmann.

PR #974, opened 2018-01-02
tipue_search: adhere to RELATIVE_URLS setting for URL generation

davidjb:
This solves an issue stemming from the changes in #591 where the generated URLs are always relative, even when using RELATIVE_URLS as False in publishing settings.  This caused URLs showing up in the Tipue Search JSON file (and thus a frontend UI) to appear like `blog/about.html` when it should be `https://example.com/blog/about.html`; in short, whilst it works, it looks strange as there is no context.

This fixes the issue by adhering to RELATIVE_URLS; if set as `True` (eg dev/testing), the existing behaviour is applied.  If set as `False` (eg for production use), then the full site URL is created using pre-#591 behaviour.

### DIFF
--- a/tipue_search/tipue_search.py
+++ b/tipue_search/tipue_search.py
@@ -30,6 +30,7 @@ class Tipue_Search_JSON_Generator(object):
         self.output_path = output_path
         self.context = context
         self.siteurl = settings.get('SITEURL')
+        self.relative_urls = settings.get('RELATIVE_URLS')
         self.tpages = settings.get('TEMPLATE_PAGES')
         self.output_path = output_path
         self.json_nodes = []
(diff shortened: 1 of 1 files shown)


jorgesumle:
-        page_url = page.url if page.url else '.'
+        if self.relative_urls:
+            page_url = page.url if page.url else '.'
+        else:
+            page_url = (self.siteurl + '/' + page.url) if page.url else '.'

The `if page.url` condition is repeated. I think it would be better if it only appears once

-        page_url = page.url if page.url else '.'
+        if self.relative_urls:
+            page_url = page.url
+        else:
+            page_url = self.siteurl + '/' + page.url

Or in one line, something like

page_url = page.url if self.relative_urls else (self.siteurl + '/' + page.url)

davidjb:
Thanks for the review - updated with a cleaner and simpler implementation.  I took a slightly different tack than your second code block as that changes the behaviour if `page.url` is falsey (the `'.'` would be overridden by later lines).

jorgesumle:
Yes, that was what I meant. It's much cleaner now.

Now it should be merged.

justinmayer:
Many thanks to David for the contribution and to Jorge for the review. 🎉

PR #978, opened 2018-01-13
Add pelican_meetup_info as submodule

tylerdave:
Add plugin that gets information about a Meetup.com group from the Meetup API and makes it available for use in pages/articles.


justinmayer:
Thanks for the contribution, Dave.

PR #979, opened 2018-01-18
[render_math] Fix Markdown headings in ReadMe 

micahjsmith:
I just add a couple spaces to get the headings to render properly


justinmayer:
Thanks for the fixes, Micah.

PR #980, opened 2018-01-27
Update jinja2content plugin

### LINKED ISSUE #809: [jinja2content] .html support?
Any chance of getting .html support for jinja2content?  It should be even easier than .rst support, right?

### LINKED ISSUE #922: [jinja2content] Feature Request: RST Support
Hello @yomguy (and anyone else with interest in this),

Per your README:

> Only Markdown supported at this moment. Adding .rst support shouldn't be too hard, and you can ask for it by opening an issue.

It'd be fantastic if you could do so. I'd love to convert to RST, but I make heavy use of this invaluable extension, so I can't just yet. Happy to help with and testing/documentation/whatever needs surrounding it. Unfortunately, Python is not my forte (yet?). Many thanks!

micahjsmith:
This is a thorough refactor of the jinja2content plugin that also adds some additional functionality.

- restructure Jinja reader as a mixin
- remove re-implementation of MarkdownReader.read
- add support for RST and HTML input through JinjaRstReader and JinjaHTMLReader
- update README to clarify functionalities

The plugin now works as follows. The mixin reads Jinja-specific settings during initialization. Then, when the reader's `read` method is called, the mixin reads the source file and renders it using the configured Jinja environment. Then, it writes the results to a temporary file and passes it on to the base reader classes. This is nice because the plugin doesn't have to make any assumptions about what the base reader class is doing, as long as it can pass a file name to the content.

This also closes #809 and closes #922.

### DIFF
--- a/jinja2content/README.md
+++ b/jinja2content/README.md
@@ -1,23 +1,31 @@
 # Jinja2 Content
 
-This plugin allows the use of Jinja2 directives inside your pelican
-articles and pages.  This template rendering is done before the final html
-is generated, i.e. before your theme's `article.html` is applied.  This
-means the context and jinja variables usually visible to your article
-template **ARE NOT** available at this time.
+This plugin allows the use of Jinja2 directives inside your pelican articles and pages.
 
-All code that needs those variables (`article`, `category`, etc) should be
-put inside the theme's template logic.  As such, the main use of this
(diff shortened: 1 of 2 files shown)


justinmayer:
Many thanks for the extensive enhancements, Micah!

PR #988, opened 2018-02-27
Add estimated read time plugin to display content read times

wayofnumbers:
An article estimated read time plugin for Pelican static site generator. After Pelican generated the content of each page, the plugin read through the generated HTML content and strip all the tags, count all the words, then utilize average human read speed to calculate the read time of each article. The read time is passed over to the 'content' object of the article so Jinja template can use it to display the read time on wherever appropriate.


jorgesumle:
@MinchinWeb coded something similiar. This may interest you: https://github.com/MinchinWeb/minchin.pelican.plugins.post_stats

wayofnumbers:
@jorgesumle Thanks for the insights. I'll definitely check it out. It seems that BeautifulSoup is used to parse the website, which require extra module installation.

justinmayer:
Many thanks for the contribution, Michael.

wayofnumbers:
Thanks @justinmayer  for the recognition! Will keep contributing while idea and time permit. :)

Issue #989, opened 2018-02-27
[subcategory] access to subcategories in templates


oulenz:
@alistairmagee

I have two pull requests that fix some things that I needed to use the subcategory plugin in my project. They change the semantics of subcategories to some extent, so I am not sure whether they fit the spirit of this plugin. The two pull request may be seen as alternatives, the second being more radical.

https://github.com/getpelican/pelican-plugins/pull/990
https://github.com/getpelican/pelican-plugins/pull/991

- The first pull requests adds `subcategories` to the context, so it can be used in templates
- It also adds categories as subcategories. Ie, categories appear twice, as categories and as subcategories. The main reason is that `subcategories` is pretty useless if it doesn't contain the top-level categories, and I didn't see what purpose the distinction between top-level categories and proper subcategories serves.
- I am not sure I see the use-case for having both categories and subcategories at all, so the second pull request simply redefines subcategories as categories.

---

There are a number of ideas that I didn't include in these pull requests but that I'll list here for context:

- Adding a setting to only include articles in their most specific subcategory
- Adding a list of children to each category to allow more fine-grained access when building menus
- Redefining the index as the unique root category (not sure whether this isn't more trouble than it's worth)

alistairmagee:
Hi Oulenz

It's been a long time since I had a look at any of this stuff.

I believe I made Categories and Subcategories separate so that existing templates would not break. ie. All templates implement categories as they normally are. On my site I only had to change the existing navigation template to

```
<nav>In:
        <ol class="category-list">
            <li>
                <a href="{{ SITEURL }}/{{ article.category.url }}">{{ article.category }}</a>
            </li>
            {% include "subcategorylist.html" %}
        </ol>
    {% include "taglist.html" %}
    </nav>
```
And subcategorylist.html is 

---
{% if article.subcategories %}{% for subcategory in article.subcategories %}
<li>
    <a href="{{ SITEURL }}/{{ subcategory.url }}">{{ subcategory.shortname }}</a>
</li>    
{% endfor %}{% endif %}
---

I thought making it easy to update existing templates to accommodate Subcategories without breaking templates for websites without them. 

I don't know what other plugins and Templates rely on the behaviour of Categories as it currently stands so I don't think its a good idea to change the way they are defined completely. So I don't think it's a good idea to merge #991.

[...]

PR #990, opened 2018-02-27
categories as top-level subcategories, make subcategories available in context

oulenz:
Fixes issue #989


oulenz:
Closed in favour of https://github.com/getpelican/pelican-plugins/pull/1082

PR #991, opened 2018-02-27
redefine subcategories as nested categories rather than separate attribute

oulenz:
Fixes issue #989


oulenz:
Closed in favour of https://github.com/getpelican/pelican-plugins/pull/1082

Issue #995, opened 2018-03-17
[asciidoc_reader] Upgrade breaks umlauts with "'ascii' codec can't decode byte 0xc3 in position 19: ordinal not in range'"


bentolor:
Upgrading from c478eb7e0dc3fdaa1e36e5829ea00cf0f0af1bb9 to 8de8e847ce78e53b62dfdcd7e0e1923614adead5 breaks the Asciidoc input for me

I have asciidoc and asciidoctor installed:

```
$ asciidoctor -v ; asciidoc --version 
Asciidoctor 1.5.6.1 [http://asciidoctor.org]
Runtime Environment (ruby 2.3.1p112 (2016-04-26) [x86_64-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:- ex:UTF-8)
asciidoc 8.6.9
```

```
Starting up Pelican and HTTP server
DEBUG: Pelican version: 3.7.1
DEBUG: Python version: 2.7.12
DEBUG: Adding current directory to system path
DEBUG: Temporarily adding PLUGIN_PATHS to system path
DEBUG: Loading plugin `asciidoc_reader`
DEBUG: Registering plugin `asciidoc_reader`
DEBUG: Restoring system path
  --- AutoReload Mode: Monitoring `content`, `theme` and `settings` for changes. ---

-> Modified: content, theme, settings. re-generating...
DEBUG: Template list: [u'!simple/archives.html', u'!simple/article.html', u'!simple/author.html', u'!simple/authors.html', u'!simple/base.html', u'!simple/categories.html', u'!simple/category.html', u'!simple/gosquared.html', u'!simple/index.html', u'!simple/page.html', u'!simple/pagination.html', u'!simple/period_archives.html', u'!simple/tag.html', u'!simple/tags.html', u'!simple/translations.html', u'analytics.html', u'archives.html', u'article.html', u'author.html', u'authors.html', u'base.html', u'categories.html', u'category.html', u'common_seo.html', u'disqus_count.html', u'gosquared.html', u'index.html', u'page.html', u'pagination.html', u'period_archives.html', u'search.html', u'tag.html', u'tags.html', u'tipue_search.html', u'translations.html']

…

DEBUG: Read file readablez/76_Schoene_Bilder.adoc -> Article
ERROR: Could not process readablez/76_Schoene_Bilder.adoc
  | UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 19: ordinal not in range(128)
  |___
[...]

PR #996, opened 2018-03-18
Fix Unicode problems reported in issue #995

jeffrimko:
Hopefully these changes will help with issue #995. I was able to test using both Python 2/3 and everything seemed okay. This change also fixes and issue seen with titles/metadata not being rendered correctly in Python 3.


justinmayer:
Thanks, Jeff!

PR #1000, opened 2018-03-23
Add support for python3 in liquid_tags/b64img.py

alexandrenorman:
The following error occured when used with python3

ImportError: No module named 'urllib2'


justinmayer:
Thanks to @alexandrenorman for the contribution and to @avaris for reviewing.

PR #1021, opened 2018-04-30
better_figures_and_images: Fix "local variable 'src' referenced before assignment"

Natim:
```
  |   File "~/git/plugins/better_figures_and_images/better_figures_and_images.py", line 77, in content_object_init
  |     if src is None:
  | UnboundLocalError: local variable 'src' referenced before assignment
```


justinmayer:
Thanks, Natim!

Natim:
You are welcome

PR #1026, opened 2018-05-06
Reddit poster

jappeace:
Posts articles to reddit, you can specify specific subs within the article by using the `subbreddit` key. You do need to enable this as an app on a reddit account, but once it's done it should make it much more easy to do publishing on reddit at least.
It just creates posts with links towards your own website.

Have fun.


justinmayer:
Thank you for the contribution, Jappie.

PR #1028, opened 2018-05-20
SiteMap: Updated sample

W7PEA:
Thanks for the plugin; ping it forward with a doc update..
The sample as-is, 'pelican.plugins.sitemap ' would not load. What I have provided works. I also added the path statement as an extra helper.


justinmayer:
Thanks, Patrick!

PR #1031, opened 2018-06-06
Fix UnboundLocalError build error

alorence:
In commit b39fe8231, a build error has been introduced. When plugin `better_figures_and_images` is used, and a `image ::` tag references an image file using `{filename}`, `|filename|` or `/static` prefix, the test `if src is None:` raises an exception:

```
ERROR: Could not process .\blog-post-file.rst
  | UnboundLocalError: local variable 'src' referenced before assignment
```

This PR simply fix this error, by defining `src` variable to None by default. I assume this was the original intention of the author


justinmayer:
I believe this was already addressed in #1021. If not, please let me know and I can re-open.

PR #953, opened 2017-10-28
[i18n_subsites] Object Draft now doesn't exist

jorgesumle:
It was removed from Pelican, so we have to use the object Article with the
value of status set to 'draft'.

This aims to fix issue #937 

I'm not able to run the tests, as I get an error I don't know how to solve. Any help is really appreciated. Ping @smartass101 and @ingwinlu

<pre><samp>
✔ ~/Código/Repositorios/pelican-plugins/i18n_subsites [fix-i18n_subsites L|✔]$ python3 test_i18n_subsites.py 
Traceback (most recent call last):
  File "test_i18n_subsites.py", line 10, in <module>
    from . import i18n_subsites as i18ns
SystemError: Parent module '' not loaded, cannot perform relative import
</samp></pre>

PR #1035, opened 2018-06-11
Fixing import in representative_image plugin - cf. #953

Lucas-C:


ashwinvis:
I confirm that this approach works in practice:

https://github.com/ashwinvis/ashwinvis.github.io/blob/73ebe3e5c97cb7422e0505ad306b575f321ac08d/pelican-plugins/representative_image/representative_image.py#L3-L10

justinmayer:
Thanks for the fix, @Lucas-C.

PR #1041, opened 2018-07-01
Fix documentation error from i18n_subsites

jorgesumle:


justinmayer:
Thanks for the fix, Jorge!

PR #1042, opened 2018-07-05
filetime_from_git: Sanitise git environment when using `git_wrapper`

cscutcher:
Otherwise peoples `~/.gitconfig` could cause commands in the wrapper to
fail.


justinmayer:
Thank you for the fix, Chris. 👍

PR #1044, opened 2018-07-08
pelican_comment_system: Link to jquery via HTTPS

pintman:
Avoiding error of mixed active content.


justinmayer:
Thanks for the fix, Marco.

PR #1048, opened 2018-07-11
Add the Similar Posts plugin

davidlesieur:
The [Similar Posts](https://github.com/davidlesieur/similar_posts) plugin adds a list of similar articles to each article's context.


justinmayer:
Thanks for the contribution, David.

PR #1050, opened 2018-07-15
Fix the yuicompressor plugin

cwfoo:
Changes:

1. Exit with error if the YUI Compressor command fails. No more silent errors.
2. Add a new setting - `YUICOMPRESSOR_EXECUTABLE` - to customize the expected name of the YUI Compressor executable.
3. Prevent problems with file names that have whitespace and shell metacharacters.
4. Remove useless verbosity option.
6. Edit README.md

Rationale for (2):

In the original readme, the suggested way of obtaining YUI Compressor is through pip. For YUI Compressor to work if it is installed using pip, the system must already have JRE installed (because YUI Compressor depends on JRE, and pip does not install JRE). For some people, this is not the best way to obtain YUI Compressor. Some people may already have YUI Compressor installed at a different path, or some may have already obtained YUI Compressor from their distribution's repositories (e.g. `sudo apt-get install yui-compressor`, which installs an executable named `yui-compressor`). Currently, the plugin expects the executable to be named `yuicompressor`, making it impossible to use copies of YUI Compressor whose executables are not named `yuicompressor`. The new `YUICOMPRESSOR_EXECUTABLE` setting allows customization of the expected executable name.

### DIFF
--- a/yuicompressor/README.md
+++ b/yuicompressor/README.md
@@ -1,18 +1,36 @@
 # YUI Compressor Plugin
 
-A pelican plugin which minify through yui compressor CSS/JS file on building step.
+A pelican plugin that minifies CSS/JS files using YUI Compressor during the
+building step.
 
 # Installation
 
-In order to work, JRE should be already installed.
-Please add `pip install yuicompressor`
+YUI Compressor needs to be present on your system. One way to obtain it is by
+installing it using pip:
 
-More info : (https://github.com/yui/yuicompressor)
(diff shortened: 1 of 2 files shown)


foochuanwei:
@auroredea @pmatos @justinmayer Do you mind taking a look? Thank you.

justinmayer:
Thank you for the contribution, @cwfoo. Apologies for the long delay!

PR #1052, opened 2018-07-18
Fix 'summary' plugin for Pelican > v3.7.1

ingwinlu:
The conditional for setting the summary is imperfect, but gets the job
done and relies on something that isn't likely to change.

Thanks to "zlg" from IRC for providing the patch.

### DIFF
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -28,7 +28,7 @@ def initialized(pelican):
 def extract_summary(instance):
     # if summary is already specified, use it
     # if there is no content, there's nothing to do
-    if hasattr(instance, '_summary'):
+    if hasattr(instance, '_summary') or 'summary' in instance.metadata:
         instance.has_summary = True
         return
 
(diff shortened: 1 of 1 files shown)


ingwinlu:
seems to need some more work

jorgesumle:
> seems to need some more work

What's wrong with that fix? It seems to work fine for another_read_more_link.

ingwinlu:
something with the replacer not being triggered.

MarkusLohmayer:
I am pretty new to pelican (so maybe this is an unrelated issue) and I have a problem related to the summary plugin. I have the 3.7.1 version of Pelican. I find that somehow - only when I have the `RELATIVE_URLS = False` -  the summary plugin stops working. I put some print() statements in `summary.py` and I came to the following conclusions:

- despite the fact that I have version 3.7.1 I always end up in the `else:` after the `if hasattr(instance, 'default_status'):`
- regardless of the value of `RELATIVE_URLS` the variable `summary` holds the right/expected content
- hence somehow setting / passing on this content is not working if `RELATIVE_URLS = False`

I would greatly appreciate help of some more experienced users of Pelican, to fix this issue. Thanks

jorgesumle:
@MarkusLohmayer, you should probably open an issue. I [solved that problem](https://github.com/fangpeishi/another_read_more_link/commit/f5a76a94036a66de96b2463c08561a79c44f3b91) for the plugin [another_read_more_link](https://github.com/fangpeishi/another_read_more_link) (the one I use) a while ago.

PR #1060, opened 2018-09-13
[liquid_tags] make docs python 2 and 3 compatible

pseyfert:
instructions were python2 only.

http://python-future.org/compatible_idioms.html#file-io-with-open


justinmayer:
Thank you for the contribution, @pseyfert.

PR #1066, opened 2018-09-21
Fix the processing of org files with SRC blocks. Fixes #1065

### LINKED ISSUE #1065: [org_reader] Error when processing org files with SRC blocks
When attempting to process `.org` files with `SRC` blocks, an error occurs.

Example org file:
```
#+TITLE: The Title Of This BlogPost
#+DATE: 2001-01-01
#+CATEGORY: blog-category
#+AUTHOR: My Name
#+PROPERTY: LANGUAGE en
#+PROPERTY: SUMMARY hello, this is the description
#+PROPERTY: SLUG first_post
#+PROPERTY: MODIFIED [2015-12-29 Di]
#+PROPERTY: TAGS my, first, tags
#+PROPERTY: SAVE_AS first_post.html

hello world! this is org-mode

#+BEGIN_SRC python
  print('hello world')
#+END_SRC
```

After running `pelican content`, the following error occurs:
```
  Please install htmlize from https://github.com/hniksic/emacs-htmlize
  ERROR: Could not process ./first_post.org
    | CalledProcessError: Command '['/usr/bin/emacs', '-Q', '--batch', '-l', '/home/kevin/projects/pelican-plugins/o
  rg_reader/org_reader.el', '--eval', '(org->pelican "/home/user/pelican/content/post.org" \'html)']' returned non-zero exit status 255.                                                        
```

This error occurs even though I have `htmlize` installed.

I believe this is because the Emacs function `org-export-as` depends on `htmlize` to process `SRC` blocks, but the `htmlize` package is not included in Emacs by default. The `-Q` command that `org_reader.py` uses calls Emacs without a config, so `htmlize` wouldn't be loaded even if installed.

kwpav:
This should fix the error indicated in Issue #1065. The `htmlize` package is needed for Emacs to process SRC blocks in org files, but it is not included in the default installation of Emacs.

It sets up the `htmlize` package for Emacs to use. I also tested this code after deleting the package -- if `htmlize` is not installed, the same error occurs as described in #1065.

I also cleaned up some of the lisp formatting, hope that is OK.

### DIFF
--- a/org_reader/org_reader.el
+++ b/org_reader/org_reader.el
@@ -1,6 +1,11 @@
 (require 'json)
 (require 'org)
 (require 'ox)
+(require 'package)
+
+;; htmlize is needed for SRC blocks
+(setq package-load-list '((htmlize t)))
+(package-initialize)
 
 (defun org->pelican (filename backend)
   (progn
@@ -14,8 +19,7 @@
             ; convert MODIFIED prop to string
             (modifiedstr (cdr (assoc-string "MODIFIED" org-file-properties t)))
             ; prepare date property
-            (dateobj (car (plist-get (org-export-get-environment) ':date)))
-            )
+            (dateobj (car (plist-get (org-export-get-environment) ':date))))
 
         ; check if #+TITLE: is given and give sensible error message if not
         (if (symbolp (car (plist-get org-export-env :title)))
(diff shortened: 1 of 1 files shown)


justinmayer:
Many thanks for the fix, Kevin!

PR #1070, opened 2018-10-07
added gitversion plugin

xunzi:
very simple plugin to include git commit id and date as template vars

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -156,6 +156,8 @@ Math Render               Gives pelican the ability to render mathematics
 
 Mbox Reader               Generate articles automatically via email, given a path to a Unix mbox
 
+More Categories           Multiple categories per article; nested categories (`foo/bar, foo/baz`)
+
 Multi Neighbors           Adds a list of newer articles and a list of older articles to every article's context.
 
 Multi parts posts         Allows you to write multi-part posts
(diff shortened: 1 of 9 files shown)


gitversion/Readme.md
```diff
@@ -0,0 +1,19 @@
+# Pelican gutversion Plugin
```
justinmayer: `gitversion` is misspelled here.

gitversion/gitversion.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+
+logger = logging.getLogger(__name__)
+
+try:
+    from git import Repo
+except ImportError:
+    logger.warning("Could not import git!")
+    sys.exit(0)
```
avaris: Uh. So... you're going to quit pelican completely because you can't import `git`? Isn't it enough to issue a warning?

PS: FWIW, this can be a simple function in the settings itself rather than a plugin, but OK I guess :).

gitversion/Readme.md
xunzi: Oh, especially stupid typo, thanks for the catch. I have added more commits to this PR, i this the correct way or should I create a new one?

justinmayer: Oh, there's definitely no reason to create a new pull request for fixing things about the original pull request. The best solution is indeed to add new commits, and once everything has been approved, [rebase & squash your commits](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits).

gitversion/gitversion.py
xunzi: @avaris Thanks for the comment, I have removed the `sys.exit`

PR #1071, opened 2018-10-19
Fix bug in ctags_generator.py

Lucas-C:
A bug in `ctags_generator/ctags_generator.py` blocked the generation of `pages` whenever this plugin was used with `WRITE_SELECTED = []` in the settings


justinmayer:
Thanks for the fix, Lucas.

Issue #1072, opened 2018-10-30
[render_math] Tries to access article.summary, which memoizes things to early


j-hui:
In `process_summary()`, `article.summary` is accessed. Because this is a memoized property, it will cause a lot of issues. In particular, if there was no summary to begin with, it will try to read from `article.content`, which is itself a memoized property, and ignore thus any subsequent changes to `article._content` will be ignored (breaking a lot of other plugins).

I'll be issuing a fix for this soon.

avaris:
What does it break exactly?

It is called in `all_generators_finalized` which is pretty late and should be safe(-ish) to access `.summary` or `.content`.

The problem with changing it to `._summary` is that you won't get math rendering in your auto-generated summaries.

j-hui:
It breaks `pelican-cite`, i believe if it is run after `render_math`.

`all_generators_finalized` is run late, but doesn't guarantee that other plugins might write to `._content` or `._summary` after `render_math` is run.

my current workaround in #1073 ignores auto-generated summaries all together. a hack is to just pull from `._content` directly and write into `._summary` if it's not already populated, but that's also risky because that might forcibly populate it too early (there might be unrendered stuff in the beginning portion of `._content` that still need to processed by other plugins)... i'm not sure there's a clean way to support that feature other than by mandating that `render_math` is the last plugin to be run

avaris:
Two suggestions:
1) `pelican-cite` doesn't need to run that late. It could easily be `content_object_init`.
2) `render_math` does a (hacky) `cache-clear` for summary. Same could be done for content.

kdeldycke:
FTR, `render_math` plugin has been moved to its own repository at: https://github.com/pelican-plugins/render-math

justinmayer:
If this issue persists, please re-submit it to the independent repository linked above.

PR #1073, opened 2018-10-30
don't access article.summary

j-hui:
fix for #1072


justinmayer:
If this issue persists, please re-submit it to: https://github.com/pelican-plugins/render-math

PR #1082, opened 2018-11-06
Add more_categories plugin

oulenz:
Plugin that allows for multiple categories per article plus nested categories (`foo/bar`, `foo/baz`)


justinmayer:
Many thanks for contributing this plugin, @oulenz. Obviously this has been a frequently-requested feature, so I'm glad it now exists as a plugin. Bravo! 🎉

PR #1084, opened 2018-11-14
i18n_subsites: Use new draft status for articles

lioman:
pelican 4 introduces `draft` as allowed status for articles and removes `Draft` from `pelican.contents`.

i18n_subsites does not work any more with version 4.x. This MR fixes this

### DIFF
--- a/i18n_subsites/i18n_subsites.py
+++ b/i18n_subsites/i18n_subsites.py
@@ -22,7 +22,10 @@ import locale
 from pelican import signals
 from pelican.generators import ArticlesGenerator, PagesGenerator
 from pelican.settings import configure_settings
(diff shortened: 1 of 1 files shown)


lioman:
Seems to work, I will update this MR later

i18n_subsites/i18n_subsites.py
```diff
@@ -153,7 +156,9 @@ def save_generator(generator):
 def article2draft(article):
     '''Transform an Article to Draft'''
     draft = Draft(article._content, article.metadata, article.settings,
+    article.status = 'draft'
                   article.source_path, article._context)
+    return article
```
avaris: Huh, what's going on here? You don't need these changes.

Issue #1083, opened 2018-11-11
[autopages] Only returns the page path, does not render


Lucidiot:
The [README](https://github.com/getpelican/pelican-plugins/blob/master/autopages/README.md) for the `autopages` plugin states the `author.page`, `category.page` and `tag.page` variables will hold the rendered content from the found pages, however they only hold a path to the pages and do not render the pages directly.

avaris:
Code looks OK to me. How are you using them in the template?

Lucidiot:
Excerpt from my `authors.html` template:

```
{% for author, articles in authors|sort %}
  ...
  {{ author.page }}
  ...
{% endfor %}
```

This prints the absolute path to the rst/md file associated to the author.

avaris:
`author.page` is an actual `Page` object: same as pelican `Page`, or very similar to `Article`. So, you use it accordingly:

```
{{ author.page.content }}
{{ author.page.summary }}
{{ author.page.some_metadata }}
# etc...
```

Lucidiot:
Oh! Okay, now it works, thanks! Shouldn't the README state that this is a Page object then, not *"rendered content"* (which I interpreted as the `.content`)?

avaris:
Probably. Want to submit a PR? :)

Lucidiot:
Here it is! #1085

Lucidiot:
Closing as #1085 is merged

PR #1085, opened 2018-11-14
[autopages] Misleading variable descriptions in README

Lucidiot:
Related to #1083: explicitly state `author.page`, `category.page` and `tag.page` variables are `Page` objects and not "rendered content" in the `autopages` README. Rendered content could be intepreted as rendered HTML, not as page objects.


justinmayer:
Thank for the clarification, @Lucidiot.

Issue #1087, opened 2018-11-21
Pelican 4.0.0 - language "th" not supported: Docutils-generated text will be in English


siongui:
I use i18n_subsites to build websites with 2 lang - en and th (Thai).
After upgrading to Pelican 4.0.0, I got the following similar messages for each processed article:

````
  | SystemMessage: /home/travis/build/siongui/ajahn-jayasaaro/content/articles/why-is-it-that-we-so-love-dhamma-by-ajahn-jayasaaro%en.rst:: (WARNING/2) language "th" not supported: Docutils-generated text will be in English.
````
[Detail log on Travis CI](https://travis-ci.org/siongui/ajahn-jayasaaro/builds/457700157)

I googled the error message and found similar issue in getnikola/nikola#2437
This issue disappears when I use Pelican 3.7.1.

**Comment by justinmayer** [comment]
In the future, if reporting an issue related to plugins (like this one), please do so in the Pelican Plugins repository — not in the Pelican core repository. Thank you.

**Comment by siongui** [comment]
> In the future, if reporting an issue related to plugins (like this one), please do so in the Pelican Plugins repository — not in the Pelican core repository. Thank you.

Thanks @justinmayer  for the advice. I open in core repo because it happens on Pelican 4.0.0 but not on 3.7.1 with the same i18n_subsites code. So I think it's related to core repo instead of plugin. Actually this is the second issue related to Pelican 4 + i18n_subsites, and the first one opened in plugin repo ( getpelican/pelican-plugins#1086 ) has been fixed.

**Comment by justinmayer** [comment]
While we try to avoid it when we can, it’s not uncommon for changes in core to break plugins. When that happens, it’s usually not an indication that something is wrong with core, but that the plugin needs to be updated for compatibility with the latest core release.

**Comment by avaris** [comment]
We may need to revert getpelican/pelican#2256 or at the very least be a bit more smart about it (i.e. don't use any language code docutils not happy about).

cc @mosra

**Comment by avaris** [comment]
@siongui As a workaround for the time being, you can use the following to override the setting:

```python
DOCUTILS_SETTINGS = {'language_code': 'en'}
```
That should make docutils not grumpy.

siongui:
@avaris the workaround added in pelicanconf.py does not work.
````
CRITICAL: KeyError: 'filenames'
````
Please refer to the build log of Travis CI for complete log.

avaris:
[...]

PR #1088, opened 2018-11-24
i18n_subsites: Accommodate filename renaming to static_content in generator context

oulenz:
Fixes https://github.com/getpelican/pelican-plugins/issues/1087 (hopefully)

The `i18n_subsites` tests are broken in multiple ways, so I've found it impossible to test whether this is backwards compatible, but it should be.


justinmayer:
Thank you for the (hopeful) fix, @oulenz. 😅

PR #1094, opened 2019-01-07
Render_Math: Fixed broken config script + Added variables for font & numbering

tabidots:
Changing `MATH_JAX[key] = value` was not affecting the final `MathJax.Config.Hub` output to the page, because the configuration script has to go _before_ the MathJax invocation, not after.

Also, it was not previously possible to force the use of TeX (or any of the other MathJax fonts) over STIX. I have fixed this, but it turned on equation numbering by default, so I have turned that off by default and made Pelican config variables for both things.

In `pelicanconf.py`, you can now specify

```python
MATH_JAX = {         'font_list': ['my_favorite_font', 'my_next_favorite'], 
            'equation_numbering': 'none/auto/AMS'}
```
where fonts can be any of `TeX, STIX-Web, Asana-Math, Neo-Euler, Gyre-Pagella, Gyre-Termes, Latin-Modern`.


justinmayer:
Thank you for the enhancement, @tabidots.

PR #1095, opened 2019-01-08
asccidoc_reader: Do not encode into latin1 on python3

gileri:
It causes UnicodeDecodeError when using non-ascii characters in titles such as 'à'.
The commented example (`ñ`) doesn't work in python3.7 with the encode/decode hack, but works fine without. I don't really understand why it was introduced.

I haven't run the test suite, I can't manage to get a working testing environment with pelican and pelican-plugins.


gileri:
Fixed in 671aaf939aa697fcc83ef51256a17fb1f8133bbe

PR #1102, opened 2019-01-23
liquid_tags: Python 3 requires from errno import EINVAL, EPIPE

cclauss:
Python 2 did not require these import but Python 3 does.


justinmayer:
Thank you for the contribution, @cclauss.

PR #1106, opened 2019-02-09
Update rmd_reader to work in Pelican 4.0

capnrefsmmat:
Upon upgrading to Pelican 4.0, I was presented with this error:

```
CRITICAL: TypeError: enabled() takes 0 positional arguments but 1 was given
```

pdb revealed this was when Pelican tried to access the `enabled` property of an `RmdReader` instance; since `enabled` was not declared to take `self`, the method doesn't work on instances. Adding the `self` argument now works perfectly.


justinmayer:
Thank you for the fix, Alex.

PR #1108, opened 2019-02-18
share_post: Remove Google Plus

askpatrickw:
Google plus is closed.


justinmayer:
Thank you for the assistance, Patrick.

PR #1109, opened 2019-02-24
interlinks: Fix images not rendering

mlabbe:
This occurs because Pelican now generates &gt; for >, so the old string replacement no longer works.

I tried to rebuild my site, but it failed, and upon investigation, it was because this string replace was failing.  &gt; did not need to be replaced with '>'


justinmayer:
Thank you for the fix, Michael.

PR #1113, opened 2019-03-02
rmd_reader: Avoid warning about the use of {filename}

jranke:
After the upgrade to pelican 4, I got warnings that told me to use {static} instead of {filename}. This change makes these warnings disappear as far as they were caused by my rmd articles.


justinmayer:
Thank you for the fix, Johannes!

PR #1115, opened 2019-03-11
better_figures_and_images: Accept the new {static} link syntax

vetetix:
This modification enables the better_figures_and_images plugin to accept the new {static} link syntax in place of the old {filename} which raises warnings when used with static content.


justinmayer:
Thank you for the fix, @vetetix.

PR #1131, opened 2019-04-22
fix 'math processing error'

tomjpsun:
Previous URL does not work well and cause this error.
( MathJax.Message.Log() reports a file cannot be loaded. )


notquitehere:
Hi,

Removing latest from the URL means that the version number has to be incremented every time it's updated (third section on here: https://www.mathjax.org/#gettingstarted), so you might want to put that back :)

T.

PR #1140, opened 2019-05-04
feature: add cookie hint on every html page (DSGVO)

ronnyfriedland:


avaris:
Wouldn't you do this in the theme?

Also, looks like this would require `jQuery` but no mention of it in the readme. 

And a defined but never used variable in the code: `JS_LIBRARIES`

ronnyfriedland:
Hi,
thank you for giving feedback.

Putting it to the theme would be much easier you are right - but you have to always use the same theme or requires modifying any theme before using.

I removed th unused vaiable and import and added the dependency to jquery.

Issue #1145, opened 2019-05-17
Issue with MathJax availableFonts setting - [Math Processing Error] 


Timothy-Pulliam:
I am having problems with my equations showing as [Math Processing Error] when I render them.

I have opened a issue with MathJax (see [here](https://github.com/mathjax/MathJax/issues/2148)) and they were able to determine the reason was, in the HTML-CSS preferences section of `mathjax_script_template`, there were some issues. I have been able to get my site working by making the following changes
```
 "    'HTML-CSS': {{ " +
 "        availableFonts: [{font_list}]," +                                                 
```

to this

```
"    'HTML-CSS': {{ " +
"        availableFonts: ['STIX', 'TeX']," +
"        preferredFont: 'STIX'," +
```
I tried to find where the error was in `math.py` but I do not see any issues, so I am unable to submit a pull request. However this appears to be a viable work around for now.

Timothy-Pulliam:
I have reopened the issue with MathJax as the issue persists despite their recomendation.

Lucas-C:
@Timothy-Pulliam : from your comment https://github.com/mathjax/MathJax/issues/2148#issuecomment-493509796 I understand you found the solution to your issue ?

moorepants:
I'm getting this on local builds of my websites, since the last month or so. It works fine when served from a remote webserver though.

Timothy-Pulliam:
This should be closed, a pull request has been approved. https://github.com/getpelican/pelican-plugins/pull/1146#event-2474676784

PR #1146, opened 2019-05-17
Fix MathJax availableFonts, prefferedFont rendering issue

Timothy-Pulliam:
I have opened an issue with MathJax for a rendering issue with render_math as a result of `preferredFont` and `availableFonts` not being set properly in MathJax.Hub.config["HTML-CSS"] section in `mathjax_script_template`. We were able to determine the following changes would fix the issue. Please see the below open issues.

https://github.com/mathjax/MathJax/issues/2148

https://github.com/getpelican/pelican-plugins/issues/1145

### DIFF
--- a/render_math/mathjax_script_template
+++ b/render_math/mathjax_script_template
@@ -33,7 +33,8 @@ if (!document.getElementById('mathjaxscript_pelican_#%@#$@#')) {{
         "        preview: '{latex_preview}'," +
         "    }}, " +
         "    'HTML-CSS': {{ " +
-        "        fonts: [{font_list}]," +
+        "        availableFonts: {font_list}," +
+        "        preferredFont: 'STIX'," +
         "        styles: {{ '.MathJax_Display, .MathJax .mo, .MathJax .mi, .MathJax .mn': {{color: '{color} ! important'}} }}," +
         "        linebreaks: {{ automatic: "+ linebreak +", width: '90% container' }}," +
         "    }}, " +


abarker:
I had the same problem with a new pull of pelican-plugins.  After reading your link above I modified `pelican-plugins/render_math/mathjax_script_template` by changing the line with `"fonts: [{font_list}]," +` to `"fonts: {font_list}," +`.  That seems to have fixed the problem, at least for now.

dvdtGravWe11:
I've just confirmed that the same issue still exists. I went through all the same debugging steps before finding this pull request. I hope they merge this fix soon. 

I'm dumping my full debugging notes here because I hope they'll be helpful to anyone checking this work.

# Summary

In `render_math/mathjax_script_template:36`, `fonts: [{font_list}]` should be `fonts: {font_list}`.

# Bug in basic install

After adding `PLUGINS = ['render_math']` to pelicanconf.py, and loading the page, math is initially rendered during MathJax's fast-render step, but then changes to `[MATH PROCESSING ERROR]` after about 30 seconds.

# Initial debugging

The javascript terminal stated the error was located in `TypeError: n.replace is not a function jax.js:19:1432`. This is minified javascript, so we need to switch to the unpacked javascript to debug.

# MathJax Source Code Debugging

Due to an unrelated issue with cdnjs, I was unable to get the unpacked version of the javascript from the cdn, so I `git clone`d the repository and ran all javascript locally. I sourced the unpacked version of the javascript so I could debug the issue.

The stack trace terminates at `mathjax/jax/jax.js:88` ([src](https://github.com/mathjax/MathJax/blob/master/unpacked/jax/output/HTML-CSS/jax.js)), with the error that variable `name` does not have a `replace()` method. Careful investigation shows that `name` is supposed to be a single string, but is actually the full array of fonts `['STIX','TeX']`.

Working up the stack, the Config class in jax.js has a property called `config.availableFonts` which is a double nested list `[['STIX','TeX']]`. `name` is being pulled from this list.

# Pelican Plugin Debugging

The pelican plugin `render_math` is setting the `fonts` variable using string replacement at `pelican-plugins/render_math/math.py:264` ([src](https://github.com/getpelican/pelican-plugins/blob/master/render_math/math.py)).

The string replacement action is performed against text file `pelican-plugins/render_math/mathjax_script_template:36` ([src](https://github.com/getpelican/pelican-plugins/blob/master/render_math/mathjax_script_template)) which I have reproduced below.

`pelican-plugins/render_math/mathjax_script_template:36`

`        "        fonts: [{font_list}]," +`
        
This substitution produces the following double nested list in the output javascript:

`        "        fonts: [['STIX', 'TeX']]," +`
        
To fix this, change `pelican-plugins/render_math/mathjax_script_template:36` to say

`        "        fonts: {font_list}," +`

paultopia:
Bump!  I just encountered the same bug and can confirm that @dvdtGravWe11 's analysis matches my own, and that this fix squishes the bug.  Pretty please merge it?

justinmayer:
@Timothy-Pulliam: Many thanks for the fix! Apologies for the delay, folks.

moorepants:
FYI, fixes this issue for me.

PR #1148, opened 2019-05-26
[more_categories] Fix category slug substitutions

oulenz:
Fixes https://github.com/getpelican/pelican/issues/2565

The slug substitution settings changed in Pelican 4, and this wasn't yet reflected in this plugin. The plugin already only worked with Pelican >= 4, so backwards compatibility is not affected.


justinmayer:
Bravo, @oulenz! Much appreciated.

PR #1150, opened 2019-05-27
Add Shortcodes plugin

Granitosaurus:
This plugin allows to define macros called shortcodes in content pages that will be expanded as jinja2 templates.
The purpose of this plugin is to allow explicit and quick jinja2 templating in your content without compromising markdown/rst.

Example:

    #pelicanconf.py

    SHORTCODES = {
        'image': "{{desc|title}}"
    }

Then in your content:

    [% image src=foo.png desc="this is a test" %]

will become:

    This Is A Test

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -248,6 +248,8 @@ Series                    Groups related articles into a series
 
 Share post                Creates share URLs of article
 
+Shortcodes                Easy and explicit inline jinja2 macros
+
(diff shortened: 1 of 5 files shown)


Granitosaurus:
Can we merge this in?

justinmayer:
Very sorry for the delay, Bernardas. This seems like an interesting plugin. Many thanks for the contribution!

PR #1154, opened 2019-06-18
i18n_subsites: Replace JINJA_EXTENSIONS with JINJA_ENVIRONMENT in docs

ivanhercaz:
The instructions about how to localize templates with Jinja2 are very confusing because if you follow the guide you would get the next error:

```
WARNING: JINJA_EXTENSIONS setting has been deprecated, moving it to JINJA_ENVIRONMENT setting.
CRITICAL: KeyError: 'JINJA_ENVIRONMENT'
```

For that I think it would be better to have it fixed and updated with the current system. The example codes are fine, they were replaced when it changed. If I am not wrong, it is the only file where this persists.


justinmayer:
Many thanks for the fix, Iván!

PR #1157, opened 2019-07-05
filetime_from_git: enable searching parent directory

talha131:
If your Pelican content is in a sub-directory then this plugin used to
fail. 

This patch fixes the issue.


justinmayer:
Thanks for the fix, Talha!

PR #1161, opened 2019-07-29
author_images: Add Python 3 compatibility

WPettersson:
Fixes python3 compatibility of the author_images plugin


justinmayer:
Many thanks for the fix, William!

Issue #425, opened 2015-03-02
PyPI package release plan for plugins?


ionelmc:
Just putting this here.

Things to care about:

* versioning
* testing (also #25)
* release management (eg: when to make releases)

ionelmc:
Also, where to actually store the code: here or individual repositories.

kylef:
Just linking this here since it's slightly related and would make this easier: https://github.com/getpelican/pelican/issues/1488

FedericoCeratto:
* Licensing (some plugins might have missing/inconsistent licenses)

FedericoCeratto:
Also see #39

justinmayer:
Anyone who wants to help make this a reality is encouraged to post a comment on this related issue: https://bitbucket.org/pypa/pypi/issues/238/request-for-new-pypi-trove-classifier-for

MinchinWeb:
It's take a bit, but [trove classifiers for PyPI](https://github.com/pypa/warehouse/issues/1650) have now been added!

moorepants:
Adding my two cents: it would be much nicer for the end user to have versioned plugins that are install-able from pypi.

justinmayer:
No promises, but hoping to rally the community and solicit help to implement this for 5.0.

nebulousdog:
I'd be so happy with modularizing this and importing individual plugins even though there would be the extra bloat of the unused projects. For years my practice has been copy-pasting pieces, which isnt maintainable.

@justinmayer you can ping me to help if you need assigning sections of the codebase etc. :D

nebulousdog:
maybe this PR (https://github.com/getpelican/pelican-plugins/pull/1215) will help get us going? I'm attempting to do the initial steps of @ionelmc's initial suggestion and allow folks to install `pelican-plugins` via `pip` and `pipenv`.

Lucas-C:
Thanks for helping on this issue !
I commented on your PR my main concerns.

moorepants:
I think the first step is to move each plugin into its own repository (if not already). I don't think it is a good idea to have a pip installable mega distribution, because you have to lock each plugin at a specific version for the collection. This will be problematic if a user wants to specify specific versions of different plugins. If you want a mega distribution you need to work out how to pin versions for each plugin in a compatible set.

nebulousdog:
ah I see @moorepants. we should do the ideal thing with so many people using this project. I'll close my issue and help with the splitting apart.

[...]

PR #1165, opened 2019-08-25
Adding muextensions

pedrohdz:
- https://github.com/pedrohdz/muextensions

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -142,3 +142,6 @@
 [submodule "similar_posts"]
 	path = similar_posts
 	url = https://github.com/davidlesieur/similar_posts.git
+[submodule "muextensions"]
+	path = muextensions
+	url = https://github.com/pedrohdz/muextensions.git
--- a/Readme.rst
+++ b/Readme.rst
@@ -156,6 +156,8 @@ Mbox Reader               Generate articles automatically via email, given a pat
 
 More Categories           Multiple categories per article; nested categories (`foo/bar, foo/baz`)
 
+muextensions              Adds *PlantUML* and *Ditaa* rst directives.  This plugin is a stand alone Python package and is best installed directly via ``pip install muextensions``.  Instructions included in the README.rst.
+
 Multi Neighbors           Adds a list of newer articles and a list of older articles to every article's context.
 
 Multi parts posts         Allows you to write multi-part posts


pedrohdz:
I reviewed `Contributing.rst` and hoping that this makes it in even though it does not conform.  :-/  The implementation of `muextensions` is more closely aligned with how plugin configuration is described in http://docs.getpelican.com/en/stable/plugins.html#how-to-use-plugins.

If this does not fly, how about adding an *Honorable mention* section to the bottom of the `Readme.rst`, and dropping the submodule?

I would love to get this out there as I think *muextensions* will be useful to others.

Many thanks!

Lucas-C:
As mentioned recently [in this PR](https://github.com/getpelican/pelican-plugins/pull/1169#issuecomment-529139584), we are trying to move away from using `git` submodules.
We encourage new plugin authors to create repos under the [pelican-plugins](https://github.com/pelican-plugins) GitHub org, or to contribute to this monolothic repo.

There is already a [PlantUML](https://github.com/getpelican/pelican-plugins/tree/master/plantuml) plugin in this repo but `muextensions` is very welcome ! It may be worth adding a short comparison table maybe.

pedrohdz:
@Lucas-C ,
Given that  `muextensions` includes support for [Hovercraft!](https://hovercraft.readthedocs.io/en/latest/) as well, adding it to the `pelican-plugins` mono-repo, or new organization does not seem to make sense, to me at least.  :-)  So I am going with option 2 ("creating a dedicated git repo and publishing it on Pypi").

More than anything I am looking for a way to make advertise it and it available for use by other _Pelican_ users.

Maybe a page like [Available Vagrant Plugins](https://github.com/hashicorp/vagrant/wiki/Available-Vagrant-Plugins) that lists related plugins?

Lucas-C:
Here iares the detailed exchange on the subject:
- https://github.com/getpelican/pelican-plugins/issues/39
- https://github.com/getpelican/pelican-plugins/issues/425

You can for example use the Pypi classifier `Framework::Pelican::Plugins` and also add it on this page: https://github.com/getpelican/pelican/wiki/Externally-hosted-plugins-and-tools

Lucas-C:
@pedrohdz What is you decision on this ? Do you plan to publish your plugins on Pypi ?
If you still wish to contribute here, it may be best to submit the code directly, not as a git sub-module but in a dedicated directory.

justinmayer:
Hey @pedrohdz! Since `muextensions` has support for projects besides Pelican, I can see your perspective. Another idea that occurred to me was moving the Pelican-specific portion into its own `pelican-plugins/muextensions` repository. That way, the plugin could list `muextensions` as a dependency, which would be installed automatically when running `pip install pelican-muextensions`. What do you think?

PR #1166, opened 2019-09-01
Fix w3c_validate plugin for Python 3

Lucas-C:
This plugin currently does not work with Python 3

This commit fixes that while staying fully backward-compatible.
It also adds a unit test for this plugin and fixes the one for the `ctags_generator` so that it will be invoked by Travis CI

### DIFF
--- a/ctags_generator/README.md
+++ b/ctags_generator/README.md
@@ -4,8 +4,7 @@ This plugin generates a `tags` file following the [CTags format](http://ctags.so
 to provide autocompletion for code editors that support it.
 
 
-Installation
-------------
+## Installation
 
 To enable, add the following to your settings.py:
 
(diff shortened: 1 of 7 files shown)


Lucas-C:
@zemanel @justinmayer : could you please review this ?

Lucas-C:
As a side question, why isn't Travis CI enabled for this repo ?
It looks like there is some configuration here already :
https://github.com/getpelican/pelican-plugins/blob/master/.travis.yml

Lucas-C:
Regarding Travis CI, I submitted another PR to help improving the situation: https://github.com/getpelican/pelican-plugins/pull/1168

justinmayer:
Since #1168 has been merged, and Travis CI enabled for this repo, let's test it out... 😉

PR #1168, opened 2019-09-03
Travis CI: fixing plugins dependencies management

Lucas-C:
As part of an effort to fix the Travis CI tests of this repository,
this commit adds a bit of dependencies management logic,
in order for all plugins to have their requirements installed during the CI pipeline execution.

So far, this does not solve all test issues, this is just a first step in the direction of having a "green CI"


Lucas-C:
Summary for Python 2.7 :
```
         top-module | success | error   | failure

   -------------------------------------------------

         libravatar |       0 |       3 |       0

   simple_footnotes |       1 |       0 |       1

       ga_page_view |       0 |       1 |       0

        liquid_tags |       1 |      11 |       0

      i18n_subsites |       8 |       0 |       1

pelican_unity_webgl |       0 |       1 |       0

  org_python_reader |       0 |       1 |       0

    collate_content |       8 |       0 |       0

        thumbnailer |       6 |       0 |       1

            summary |       4 |       0 |       0

         gzip_cache |       3 |       1 |       1

      similar_posts |       0 |       1 |       0

               nose |       0 |       4 |       0

    more_categories |       3 |       0 |       0

    multi_neighbors |       0 |       1 |       0

      image_process |       1 |       4 |       1

                pdf |       0 |       1 |       0

         post_stats |       0 |       1 |       0

        jpeg_reader |       0 |       1 |       0

----------------------------------------------------------------------

Ran 76 tests in 5.499s

FAILED (SKIP=9, errors=31, failures=5)
```

.travis.yml
```diff
@@ -1,13 +1,18 @@
...
-    - pip install -e git://github.com/getpelican/pelican.git#egg=pelican
-    - pip install --use-mirrors Markdown
-    - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install --use-mirrors webassets cssmin; fi
-script: nosetests
+    - pip install -r requirements.txt
+    # Installing pelican in editable mode makes its tests.support package available to us
+    - pip install -e git+https://github.com/getpelican/pelican.git#egg=pelican
+    # The following commands install all plugins dependencies.
+    # The `rm`, `grep` & `sed commands` below are needed because some submodules have conflicting requirements.
+    # Also, Cython is required for the plugins pelican-ipynb through jupyter & panorama through pandas.
```
avaris: Why does `Cython` need to be explicit? Doesn't it resolve through `jupyter` or `pandas`?

pelican_unity_webgl/__init__.py
```diff
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
 
-from .UnityGameDirective import register
+from .UnityGameDirective import *
```
avaris: Why?

show_source/show_source.py
```diff
@@ -1,6 +1,9 @@
 import os
 import logging
-from urlparse import urljoin
+try:
+    from urllib.parse import urljoin
+except ImportError:  # Python 2 fallback:
+    from urlparse import urljoin
```
avaris: Since it's pretty safe to assume that `pelican` is installed and `six` is a dependency of it:
```python
from six.moves.urllib.parse import urljoin
```

liquid_tags/requirements.txt
```diff
@@ -0,0 +1 @@
+pytest
```
avaris: Can `nose` run `py.test` tests?

show_source/show_source.py
Lucas-C: Ok, fixed

liquid_tags/requirements.txt
Lucas-C: Not the ones defined as simple functions no, but the standard ones defined as classes yes: https://github.com/getpelican/pelican-plugins/blob/master/liquid_tags/test_generation.py

At some point it may be interesting to switch from `nose` to `pytest` as the former one is not much maintained anymore, but it should probably be done in another PR

.travis.yml
Lucas-C: My mistake, this is not needed indeed

pelican_unity_webgl/__init__.py
Lucas-C: My mistake, this is not needed indeed

.travis.yml
Lucas-C: Without it, this is the kind of errors I get:
https://travis-ci.org/Lucas-C/pelican-plugins/jobs/580703256
```
   Exception: Cython-generated file 'pandas/lib.c' not found.
                    Cython is required to compile pandas from a development branch.
                    Please install Cython or download a release package of pandas.
```

Lucas-C: My mistake, this is not needed indeed
This error was due to the locked PyYaml dependency from the `panorama` plugin

Lucas-C:
The follow up of this work will be done in https://github.com/getpelican/pelican-plugins/issues/1170

Lucas-C:
Note: given the current [Python 2 retirement countdown status](https://pythonclock.org), I dropped the support for Python 2 in this commit. This will also make fixing the CI pipeline easier.

liquid_tags/requirements.txt
avaris: Thanks :+1:

avaris:
> Note: given the current Python 2 retirement countdown status, I dropped the support for Python 2 in this commit. This will also make fixing the CI pipeline easier.

Can you elaborate this?

If that means making plugins Python3-only, that would be a bit premature before `pelican` itself drops Python2 support.

Lucas-C:
No, this just mean no automated tests will be run for Python 2 for this repo.
If we manage to fix the tests, the situation will still be far better than now, with zero green CI pipeline.
We will still be able to restore the Python 2 target in `.travis.yml` once the tests will all pass for Python 3.

justinmayer:
@avaris: Any further comments on @Lucas-C's plugin dependency management improvements here?

justinmayer:
Many thanks to Lucas for the improvements and to @Avaris for reviewing.

Issue #130, opened 2014-01-16
Explicit policy about submodule contributing


JocelynDelalande:
This repository is a mix of submodules and actual folders.

Contributing.rst does only mention folders. What is the contribution policy about submodules ?

Lucas-C:
Related to https://github.com/getpelican/pelican-plugins/issues/39 & https://github.com/getpelican/pelican-plugins/issues/213

We now encourage dedicated repositories, with packages published on Pypi, instead of submmodules

justinmayer:
To clarify a bit...

We hope to soon get to a point where Pelican plugins are Pip-installable and hosted on PyPI. Moreover, we encourage plugin authors to request a repository for their plugin under the new [pelican-plugins](https://github.com/pelican-plugins) GitHub organization. That way, the Pelican Dev Team, as well as other members of the community, will be able to assist with issues, code review, and managing pull requests. To see a model for how this would work, please have a look at the [Jazzband](https://github.com/jazzband/) GitHub organization.

This has been discussed previously (see: https://github.com/getpelican/pelican/issues/1488 and https://github.com/getpelican/pelican/pull/1914) but has yet to be enshrined in our documentation. Once we start actually moving existing (and new) plugins and themes to the respective new organizations, I'm sure we'll have more canonical instructions and guidance in place.

In the interim, I think the best approach is to continue with the current monolithic repository. Hopefully we'll make more progress on this endeavor in the months to come.

justinmayer:
Pelican plugins are indeed now Pip-installable as of Pelican 4.5+, so if you would like to help migrate legacy plugins to the new format, please create an issue at the [legacy monolithic Pelican Plugins repository](https://github.com/getpelican/pelican-plugins) and ask a maintainer to create a corresponding new repository under the new [Pelican Plugins organization](https://github.com/pelican-plugins) and invite you to join it.

PR #1169, opened 2019-09-03
Add Webring Plugin

davidag:
This plugin retrieves the latest posts from a list of web feeds and makes them available in templates, effectively creating a [webring](https://en.wikipedia.org/wiki/Webring). Posts are sorted from newer to older.

It is inspired in [openring](https://git.sr.ht/~sircmpwn/openring), a tool for generating an html file to include in your Static Site Generator from a template and a list of web feeds.

This is an example of the plugin in use:

example


davidag:
@Lucas-C I wasn't aware of the dedicated repositories policy that you just mentioned in #130, I only read the docs and this looked like the way to go :sob: 

In any case, it's perfectly OK for me to close this PR and follow the new guidelines when available :relaxed: 

Thank you and great work with the plugins repo! :tada:

Lucas-C:
Hi @davidag !

There are 3 ways to create plugins:
- by adding a directory to this repo
- by creating a dedicated git repo and publishing it on Pypi
- and finally by creating a dedicated git repo and referencing it here as a submodule

I think the consensus is to avoid this last approach from now on.
However the other 2 options are totally fine for me. I personnally even prefer the "collective repo" approach as it potentially allow for more code reuse & code reviews.
As you want really :)

justinmayer:
As I mentioned in my [related comment](https://github.com/getpelican/pelican-plugins/issues/130#issuecomment-529140401), I think the best approach for now is to continue with the current monolithic repository (option 1 above).

webring/README.md
```diff
@@ -0,0 +1,116 @@
+# Webring Plugin
+
+This plugin retrieves the latest posts from a list of web feeds and makes
+them available in templates, effectively creating a [partial webring]. Posts
```
justinmayer: I think this was supposed to be a link, which presumably means it should instead be: `[partial webring][]`

webring/README.md
```diff
@@ -0,0 +1,116 @@
...
+are sorted from newer to older.
+
+It is inspired in [openring](https://git.sr.ht/~sircmpwn/openring), a tool for
```
justinmayer: This should instead be: "It is inspired _by_ [...]"

webring/README.md
```diff
@@ -0,0 +1,116 @@
...
+generating an html file to include in your [SSG] from a template and a list of
```
justinmayer: html → HTML. Also, to be a link, the `[SSG]` needs to be changed to `[SSG][]` (same issue as above).

webring/README.md
davidag: In fact it's a [reference link], specifically a `link label` :open_mouth: 

[reference link]: https://github.github.com/gfm/#reference-link "This is a reference link, too!"

webring/README.md
davidag: Fixed html lowercase. The link is also _reference link_.

webring/README.md
justinmayer: Hmm. Not familiar with that bit of (presumably) GitHub-specific Markdown syntax. Not sure I see the benefit of deviating from standard Markdown just to save two characters, but perhaps I'm misunderstanding something. 🤷‍♂️

davidag: I was the one misunderstanding something :disappointed:

justinmayer:
Nice work, David. Many thanks for the cool addition to the plugin ecosystem!

Issue #213, opened 2014-05-25
Allow installing with pip


mrfuxi:
It would be great if plugins could be installed with pip

lioman:
It could be encouraged by only letting pip installable plugins into the catalog

moorepants:
Agreed, it is so much easier to manage if plugins are all installable by package managers. It would enable pinning plugin versions, dependency resolution, remove the need to clone a giant repository of many plugins, and just make it much easier to manage plugins (ant themes!).

PR #1171, opened 2019-09-04
Using a fork of image process plugin repo

Lucas-C:


MinchinWeb:
I'm not opposed to this in the slightest, but note that the internal path structure (to the plugin itself) is different between the two forks, so this will be a breaking change for anyone using the plugin. I *think* the needed change will be to adjust your plugin name from `image_process` to `image_process.minchin.pelican.plugins.image_process`.

As the current version doesn't support Pelican 4, the change is probably worthwhile none the less.

Ultimately, I think the plugins should be available on PyPI, but I realize we're not there yet.

Lucas-C:
Thank you @MinchinWeb !

About Pypi-based plugin management, _cf._ https://github.com/getpelican/pelican-plugins/issues/213

This PR is pending resolution of https://github.com/whiskyechobravo/image_process/issues/18

PR #1172, opened 2019-09-04
Fix gzip_cache unit tests

Lucas-C:
Tested with:

    nosetests gzip_cache


justinmayer:

PR #1173, opened 2019-09-05
Fix: test_simple_footnotes.py

Lucas-C:
Tested with:

    nosetests simple_footnotes


justinmayer:

PR #1174, opened 2019-09-05
Fix i18n_subsites tests

Lucas-C:
Tested with:

    nosetests i18n_subsites


avaris:
This is all updating expected outputs to newest `pelican`, right? Or did I miss something?

I guess it is alright :). I don't really know what's the expected output.

Lucas-C:
Yes, I just regenerated the expected output. I admit I did not dig much deeper than checking that the generated website can be displayed in my browser.

justinmayer:
Many thanks to @Lucas-C for the new plugin and to @avaris for reviewing. ✨

PR #1175, opened 2019-09-05
Fix JPEG reader tests

Lucas-C:
Tested with:

    nosetests jpeg_reader


justinmayer:

PR #1176, opened 2019-09-05
Add test for plantuml plugin

Lucas-C:


justinmayer:
Thank you for the new test, Lucas!

PR #1177, opened 2019-09-05
Fixing libravatar plugin tests

Lucas-C:
Tested with:

    nosetests libravatar


justinmayer:
Thank you for the fix, Lucas!

PR #1179, opened 2019-09-05
Fix sub_parts plugin tests

Lucas-C:
Tested with:

    nosetests sub_parts


justinmayer:
Thank you for the fix, Lucas!

PR #1180, opened 2019-09-05
Fix tag_cloud plugin tests

Lucas-C:
Tested with:

    nosetests tag_cloud


justinmayer:
Thank you for the fix, Lucas!

PR #1181, opened 2019-09-05
Fix thumbnailer plugin tests

Lucas-C:
Tested with:

    nosetests thumbnailer


justinmayer:
Thank you for the fix, Lucas!

PR #1183, opened 2019-09-05
Fix photos plugin tests

Lucas-C:
Tested with:

    nosetests photos


justinmayer:
Thank you for the fix, Lucas!

PR #1188, opened 2019-09-10
Add tests for render math

Lucas-C:


justinmayer:
Thank you for the fix, Lucas!

PR #1197, opened 2019-09-16
Fix tests failing due to OrgEmacsReader constructor too early check for ORG_READER_EMACS_LOCATION setting

Lucas-C:


justinmayer:
Thank you for the fix, Lucas!

PR #1199, opened 2019-09-16
optimize files - html, xml, js, json, css and svg compression using m…

pouya-abbassi:
…inify

### DIFF
--- a/optimize_files/README.md
+++ b/optimize_files/README.md
@@ -1,0 +1,24 @@
+Optimize HTML, XML, js, css and SVG
+==================================
+
+This plugin uses [minify](https://minifierr.org) to compress files.
+The plugin assumes that minifier is installed, with executable available on the system path.
+[Demo blog](https://pouyacode.net) optimized using this plugin.
+
+Installation
+------------
+
+To enable, ensure that `optimize_file.py` is put somewhere that is accessible.
+Then use as follows by adding the following to your settings.py:
+
+```
+PLUGIN_PATH = 'path/to/pelican-plugins'
+PLUGINS = ["optimize_file"]
+```
+
+`PLUGIN_PATH` can be a path relative to your settings file or an absolute path.
+
+Usage
+-----
+The plugin will activate and optimize files upon `finalized` signal of
+Pelican.
--- a/optimize_files/optimize_files.py
+++ b/optimize_files/optimize_files.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+
+"""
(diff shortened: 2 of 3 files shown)


Lucas-C:
Hi !
Thank you for contributing this plugin.
A few comments:
- as your plugin relies on the [minify](https://github.com/matthiasmullie/minify) PHP program, could you maybe rename it `minify`, and also add a check in your code that the command `minify` is available ?
- could you make `COMMANDS` configurable ? It should only be a matter of calling: `commands = pelican.settings.get('MINIFY_COMMANDS', DEFAULT_COMMANDS)`
- could you please add a least one test ?

Lucas-C:
As an extra question, would di you install `minify` ? Through `composer` ?
Because we could have it in our Travis CI if it's needed by your tests.

optimize_files/README.md
```diff
@@ -0,0 +1,24 @@
+Optimize HTML, XML, js, css and SVG
+==================================
+
+This plugin uses [minify](https://minifierr.org) to compress files.
```
avaris: Typo in URL

avaris: Well, I'm not sure if it's a typo or a completely different minifier was intended here. The one in `minifier.org` only claims to minify `JS` and `CSS`.

optimize_files/optimize_files.py
```diff
@@ -0,0 +1,51 @@
...
+    :param name: A file name to be optimized
+    """
+    filepath = os.path.join(dirpath, filename)
+    logger.info('optimizing %s', filepath)
+
+    ext = os.path.splitext(filename)[1]
+    command, silent, verbose = COMMANDS[ext]
+    flags = verbose if SHOW_OUTPUT else silent
+    command = command.format(filename=filepath, flags=flags)
+    call(command, shell=True)
```
avaris: Please prefer the `command` as a list of strings with `shell=False`.

optimize_files/optimize_files.py
```diff
@@ -0,0 +1,51 @@
...
+from pelican import signals
+
+logger = logging.getLogger(__name__)
+
+# A list of file types with their respective commands
+COMMANDS = {
+    # '.ext': ('command {flags} {filename', 'silent_flag', 'verbose_flag')
+    '.html': ('minify "{filename}" -o "{filename}"', '--quiet', ''),
+    '.xml': ('minify "{filename}" -o "{filename}"', '', ''),
+}
```
avaris: Shouldn't there be `{flags}` somewhere here?

PR #1200, opened 2019-09-21
Fix w3c_validate plugin tests

Lucas-C:
Tested with:

    nosetests w3c_validate


justinmayer:
Thank you for the fix, Lucas.

PR #1201, opened 2019-09-21
Fix TypeError in tests: expected str, bytes or os.PathLike object, not NoneType

Lucas-C:
Tested with:
```
nosetests ctags_generator
nosetests more_categories
nosetests neighbors
nosetests render_math
nosetests shaarli_poster
nosetests sub_parts
nosetests tag_cloud
```


Lucas-C:
It was due to `output_path=None` being passed to `ArticlesGenerator`
Strangely it happens in Travis CI but not on my laptop

justinmayer:
Thank you for the fixes, Lucas.

PR #1207, opened 2019-09-23
Fix tests for w3c_validate plugin

Lucas-C:


justinmayer:
@Lucas-C: Do you know why the ellipsis and curly quotation mark cause this test to fail? Should we replace them with `&hellip;` and `&rsquo;` HTML entities instead?

Lucas-C:
Honestly I'm not sure. On my laptop the test pass but on Travis CI this error is raised:
```
  File "/home/travis/build/getpelican/pelican-plugins/w3c_validate/test_wc3_validate.py", line 16, in test_generate_ctags
    validate_files(PelicanMock({'OUTPUT_PATH': TEST_CONTENT_DIR}))
  File "/home/travis/build/getpelican/pelican-plugins/w3c_validate/wc3_validate.py", line 25, in validate_files
    validate(filepath)
  File "/home/travis/build/getpelican/pelican-plugins/w3c_validate/wc3_validate.py", line 45, in validate
    vld.validate_file(filename)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/py_w3c/validators/html/validator.py", line 58, in validate_file
    content = w.read()
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 2277: ordinal not in range(128)
```
_cf._ https://travis-ci.org/getpelican/pelican-plugins/builds/588682421

avaris:
travis fails likely because it doesn't have a `utf-8` like locale set.

Python [`open` docs](https://docs.python.org/3.7/library/functions.html#open) will say that the default `encoding` is system dependent and obtained with `locale.getpreferredencoding()`.

Unfortunately, since the error comes from a third party library the plugin is depending, no easy fix besides setting a `utf-8` supported locale on travis or eliminating non-ascii characters.

```
$ LC_ALL=C python -c "import locale; print(locale.getpreferredencoding())"
ANSI_X3.4-1968

$ LC_ALL=C.UTF-8 python -c "import locale; print(locale.getpreferredencoding())"
UTF-8

$ LC_ALL=en_US.utf8 python -c "import locale; print(locale.getpreferredencoding())"
UTF-8
```

Lucas-C:
Thanks for the explanation @avaris 
Defining `LC_ALL=C.UTF-8` in Travis did not solve the issue : https://travis-ci.org/getpelican/pelican-plugins/builds/588843379
Would you have another idea on how to solve this ?

avaris:
Can you add these to `before_install` so that we can see what's going on?

```
locale -a
locale
python -c "import locale; print(locale.getpreferredencoding())"
```

Lucas-C:
I didn't know about Travis YAML `env.global`, thanks for the tip !

Pipeline output: https://travis-ci.org/getpelican/pelican-plugins/builds/588880139?utm_source=github_status&utm_medium=notification

avaris:
This is both interesting and confusing at the same time. travis seems to be fine even without fiddling with `LC_ALL` (set to `en_US.UTF-8` previously). Some other test could be messing with locale. I'm very suspicious of [this one](https://github.com/getpelican/pelican-plugins/blob/master/rmd_reader/rmd_reader.py#L37-L39).

justinmayer:
Very odd, indeed.

@wilsonfreitas: Hi! Long time no see. Could you perhaps explain the rationale behind the [code that changes the locale](https://github.com/getpelican/pelican-plugins/blob/master/rmd_reader/rmd_reader.py#L37-L39) in the RMD reader plugin?

Lucas-C:
I was able to install `R` and this repo on a machine,
and dropped a `import locale; print(locale.getpreferredencoding())` just after https://github.com/getpelican/pelican-plugins/blob/master/rmd_reader/rmd_reader.py#L37-L39
`UTF-8` was printed, so I do not think those `R` commands "leak" in the Python process

wilsonfreitas:
Hi guys,
Honestly, I don't remember why the hell the LOCALE is called this way.
Unfortunately, I can't work on it now, so feel free to let it commented if this is causing problems.
Maybe nowadays this isn't necessary.

Lucas-C:
This job execution shows however that excluding the `rmd_reader` plugin from the tests make the `w3c_validate` test pass: https://travis-ci.org/getpelican/pelican-plugins/builds/589149003
So there is probably a side-effect indeed here...

@avaris @justinmayer : I see 2 options here:
- exclude the `rmd_reader` plugin from the Travis CI pipeline, as long as no-one volunteers to fix it
- try to remove those `Sys.setlocale` lines from `rmd_reader.py`

What do you think ?

justinmayer:
Since Wilson doesn't know why those locale-setting lines are present, I suggest we remove them and see what happens.

Lucas-C:
Seems that it solved the issue :) This PR can be merged

justinmayer:
Many thanks to Lucas for the fix and to @avaris and Wilson for the feedback.

PR #1208, opened 2019-09-23
Fix liquid tags tests

Lucas-C:


justinmayer:
Thanks for the fix, Lucas. Could you elaborate as to why the `safe=True` arguments had to be removed?

Lucas-C:
It was to avoid this error I caught in [ArticleGenerator.generate_context()](https://github.com/getpelican/pelican/blob/master/pelican/generators.py#L627):

    TypeError("store() got an unexpected keyword argument 'safe'")

Lucas-C:
I tracked this change to this commit on the `markdown` lib:
https://github.com/Python-Markdown/markdown/commit/7f63b20b819b83afef0ddadc2e210ddce32a2be3#diff-0f2fbb4d1e47771e120a801756331eceL144

justinmayer:
Ah, okay — got it. Thank you for the fix and the explanation.

As a general practice, let's try to record the rationale for changes in the commit description. Otherwise, years from now, it becomes harder to recall why the changes were made. 😅 

For more info on this philosophy, see this thread: https://twitter.com/r00k/status/1175100703829909505

PR #1209, opened 2019-09-24
Fix photos plugin tests by excluding pandoc_reader

Lucas-C:
This plugin reader caused urlescaping of curly braces,
which broke tests "test_filename_article_body" & "test_photo_article_body"


justinmayer:
Thanks for the fix, Lucas.

PR #1211, opened 2019-09-25
Summary: Replace regex tag stripping with BeautifulSoup parsing

### LINKED ISSUE #1203: [summary] code blocks are messed up in article summary
Using a current Pelican version and `pelican-plugins` checkout, rendering code blocks in the summary of an article fails to render correctly:

render-bad

It appears the source of this is in #758 which indiscriminately removes all `<div>` tags, including the `<div class="highlight">` necessary for code blocks. Removing the two lines that strip those tags results in the expected summary content:

render-correct

Going through the generated output after reverting that PR, I can't find any obvious trailing `<div>` tags that the PR fixes, so possibly that problem has gone away in the intervening years.

edelooff:
This resolves #1203, ensuring well-formed markup (the intent of #758).

### DIFF
--- a/summary/Readme.rst
+++ b/summary/Readme.rst
@@ -1,3 +1,10 @@
+Requirements
+------------
+
+This plugin requires beautifulsoup4 to ensure correctness of the extracted markup::
+
+    pip install pillow beautifulsoup4
+
 Summary
 -------
 
--- a/summary/Readme.rst
+++ b/summary/Readme.rst
@@ -3,7 +3,7 @@ Requirements
 
 This plugin requires beautifulsoup4 to ensure correctness of the extracted markup::
 
-    pip install pillow beautifulsoup4
+    pip install beautifulsoup4
 
 Summary
 -------
--- a/summary/requirements.txt
+++ b/summary/requirements.txt
@@ -1,0 +1,1 @@
+beautifulsoup4
--- a/summary/test_summary.py
+++ b/summary/test_summary.py
@@ -91,6 +91,20 @@ class TestSummary(unittest.TestCase):
         self.assertEqual(page.summary, TEST_SUMMARY)
         self.assertEqual(page.content, '<p>' + TEST_SUMMARY + '</p>' + TEST_CONTENT)
 
+    def test_correct_malformed_markup(self):
+        page_kwargs = self._copy_page_kwargs()
+        del page_kwargs['metadata']['summary']
+        malformed = '<article><div><h2>Title</h2><p>Some content</article>'
+        wellformed = (
+            '<article><div><h2>Title</h2>'
+            '<p>Some content</p></div></article>')
+        page_kwargs['content'] = (
+            '<!-- PELICAN_BEGIN_SUMMARY -->' + malformed +
+            '<!-- PELICAN_END_SUMMARY -->')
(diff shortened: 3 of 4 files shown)


Lucas-C:
Could you please add a `requirements.txt` mentioning `beautifulsoup4` in the `summary/` plugin directory please ?
And also maybe a test ?

edelooff:
@Lucas-C I've added a `requirements.txt` and put a brief mention of the requirements in the `Readme.rst`. I've also attached a test that verifies that malformed content is actually corrected. 

I'm *fairly* certain this is the HTML5-correct approach, but it's proving difficult to find a quick reference to that. It's certainly the behavior that Firefox and Chrome mimic, and it's the output generated from beautifulsoup, so there's at least some authority to it.

edelooff:
For what it's worth, this is what the html5 spec has to say on (what I'm fairly sure is) this case:

[12.2.6.4.7 The "in body" insertion mode](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inbody)

**Any other end tag**
Run these steps:

1. Initialize _node_ to be the current node (the bottommost node of the stack).
2. _Loop_: If _node_ is an HTML element with the same tag name as the token, then:
   1. Generate implied end tags, except for HTML elements with the same tag name as the token.
   2. If node is not the current node, then this is a parse error.
   3. Pop all the nodes from the current node up to _node_, including _node_, then stop these steps.
3. Otherwise, if node is in the special category, then this is a `parse error`; ignore the token, and return.
4. Set _node_ to the previous entry in the `stack of open elements`.
5. Return to the step labeled _Loop_.

summary/requirements.txt
```diff
@@ -0,0 +1 @@
+beautifulsoup4
```
Lucas-C: 👍

summary/Readme.rst
```diff
@@ -1,3 +1,10 @@
+Requirements
+------------
+
+This plugin requires beautifulsoup4 to ensure correctness of the extracted markup::
+
+    pip install pillow beautifulsoup4
```
Lucas-C: I do not think `pillow` is needed.

summary/test_summary.py
```diff
@@ -91,6 +91,20 @@ def test_use_first_paragraph(self):
         self.assertEqual(page.summary, TEST_SUMMARY)
         self.assertEqual(page.content, '<p>' + TEST_SUMMARY + '</p>' + TEST_CONTENT)
 
+    def test_correct_malformed_markup(self):
```
Lucas-C: 👍 Thank you !

summary/Readme.rst
edelooff: It totally isn't, copy/paste failure. I'll fix that when I get a moment later today

avaris:
@edelooff bs4's authority goes as far as the parser it uses :). See [this](https://beautiful-soup-4.readthedocs.io/en/latest/#installing-a-parser) and [this](https://beautiful-soup-4.readthedocs.io/en/latest/#differences-between-parsers) if you like subtleties.

justinmayer:
Many thanks to @edelooff for the fix and to @Lucas-C and @avaris for reviewing. ✨

PR #1213, opened 2019-09-28
Fix priority order for math processing (fixes #1212)

### LINKED ISSUE #1212: Math uses wrong delimiters in display mode ($$) with render_math
Using the current version of `render_math` and Pelican 4.1.2, inline math is rendered correctly (i.e. `$math$` is converted to `\(math\)`, which is then rendered with MathJax), but display mode math gets converted to `\($math$\)`, and MathJax renders the $.

Here's minimal example using Markdown:

```
Title: Math time
Date: 2019-09-27 09:30

Here's some inline math: $x + y = 30$

And here's some display math:

$$
x + y = 30
$$
```

This gets converted to the following HTML:

```html
<p>Here's some inline math: <span class="math">\(x + y = 30\)</span></p>
<p>And here's some display math:</p>
<p><span class="math">\($
x + y = 30
$\)</span></p>
```

andrewheiss:
In issue #1212, inline math gets processed first, leaving $s in display math and rendering it incorrectly. Switching the priority order here makes it so that the double $$ for display math gets processed first.

### DIFF
--- a/render_math/pelican_mathjax_markdown_extension.py
+++ b/render_math/pelican_mathjax_markdown_extension.py
@@ -146,8 +146,8 @@ class PelicanMathJaxExtension(markdown.Extension):
         # Process mathjax before escapes are processed since escape processing will
         # intefer with mathjax. The order in which the displayed and inlined math
         # is registered below matters: we should have higher priority than 'escape' which has 180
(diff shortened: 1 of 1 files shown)


justinmayer:
Many thanks to @andrewheiss for the fix and to @Lucas-C and @avaris for reviewing. ✨

Issue #1189, opened 2019-09-10
Display a warning for git submodules repos installable with pip


Lucas-C:
The following repos from [.gitmodules](https://github.com/getpelican/pelican-plugins/blob/master/.gitmodules) can be installed with `pip`:
- https://github.com/kura/pelican-fontawesome.git
- https://github.com/kura/pelican_youtube.git
- https://github.com/kura/pelican_vimeo.git
- https://github.com/streeter/pelican-gist.git
- https://github.com/bmcorser/pelicanfly.git
- https://github.com/La0/pelican-flickr.git
- https://github.com/kura/pelican-githubprojects.git
- https://github.com/TC01/pelican-mboxreader
- https://github.com/tylerdave/pelican-meetup-info
- https://github.com/joachimneu/pelican-jinja2content.git -> deprecated in favour of a new version hosted in this repo

We should find a way to display a deprecation warning to users of this repo to encourage them to switch to the `pip install $plugin` method of installation.

avaris:
If the authors want those plugins to be `pip`-only, then they should remove submodules. We can extend readme with a list of "`pip`-installable-plugins". Otherwise, I don't see the point of a deprecation warning.

Lucas-C:
@avaris: My end goal is to remove git submodules from this repo for plugins that can be installed with `pip`. In order to do that, I think we should warn our users.

Lucas-C:
In fact, in the list above, only `pelican-meetup-info` & `pelican-jinja2content` have an `__init__.py` file and can be imported use through this repo and its git submodules.

I thinks we can simply remove all the other repos in this list from `pelican-plugins` git submodules. Would you agree with that @avaris ?

Also, for `pelican-meetup-info`, I tested having both `pelican-plugins` configured and doing `pip install pelican-meetup-info`: when using the plugin, the one in `pelican-plugins` is used instead of the one installed with `pip` :(

Lucas-C:
I submitted this PR in order for `pelican-plugins` plugins to be overriden by the ones from Pypi: https://github.com/getpelican/pelican/pull/2617

avaris:
I feel like the decision to "keep submodule or remove it in favor of `pip`" should be left to the author of the plugin.

> Also, for pelican-meetup-info, I tested having both pelican-plugins configured and doing pip install pelican-meetup-info: when using the plugin, the one in pelican-plugins is used instead of the one installed with pip :(

[...]

PR #1215, opened 2019-10-01
add setup config to install via pip and pipenv

nebulousdog:
hoping to relieve the following problem by adding the minimum of a `setup.py` config. if this works, will also be a good launching point for the issue at https://github.com/getpelican/pelican-plugins/issues/425.

```
> pipenv install git+ssh://github.com/getpelican/pelican-plugins
Installing git+ssh://github.com/getpelican/pelican-plugins…
WARNING: pipenv requires an #egg fragment for version controlled dependencies. Please install remote dependency in the form git+ssh://github.com/getpelican/pelican-plugins#egg=<package-name>.
✘ Installation Failed 
```

Things that I could use input from primary contributors:

1. even though this project has been around for a while, there aren't any releases listed in the [github releases](https://github.com/getpelican/pelican-plugins/releases) tab, would you like to start at semver `0.0.1`?
2. I referenced `pelican-plugins`'s [LICENSE](https://github.com/getpelican/pelican-plugins/blob/master/LICENSE) file for equivalent pypi [classifier](https://pypi.org/classifiers/) string. If you'd like this to be 3+ etc, or instead link directly to the file or such that may be important to you.
3. I'm not qualified to add authors etc. Please take those portions over during this PR or after. 

Things to do after this:

* register project with pypi via twine
  * https://pypi.org/project/twine/
* while youre at it, whomever has that token/creds, make a PR for automatic releases on git tags (or whatever criteria you want).
  * https://docs.travis-ci.com/user/deployment/pypi/
* after this is done, can change the instructions from referencing the github repo url, to the registered pypi project name
  * https://pip.pypa.io/en/stable/user_guide/#requirements-files

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,20 @@
+"""minimum config to enable distribution via pip and pipenv."""
+
+import setuptools
+
+with open('README.rst', 'r') as fh:
+    LONG_DESCRIPTION = fh.read()
+
+setuptools.setup(
+    name='pelican-plugins',
+    version='0.0.1',
+    description='A collection of plugins for the Pelican static site generator.',
+    long_description=LONG_DESCRIPTION,
+    long_description_content_type='text/x-rst',
+    url='https://github.com/getpelican/pelican-plugins',
+    packages=setuptools.find_packages(),
+    license='License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
+    classifiers=[
+        'Framework :: Pelican :: Plugins',
+    ],
+)
--- a/Readme.rst
+++ b/Readme.rst
@@ -16,6 +16,14 @@ The easiest way to install and use these plugins is to clone this repo::
 
     git clone --recursive https://github.com/getpelican/pelican-plugins
 
+Or via Pip::
+
+    pip install git+https://github.com/getpelican/pelican-plugins
+
+Of course equivalently via `pipenv <https://github.com/pypa/pipenv>`_::
+
+    pipenv install git+https://github.com/getpelican/pelican-plugins
+
 and activate the ones you want in your settings file::
 
     PLUGIN_PATHS = ['path/to/pelican-plugins']
@@ -233,7 +241,7 @@ Random article            Generates a html file which redirect to a random artic
 
 Read More link            Inserts an inline "read more" or "continue" link into the last html element of the object summary
 
-Readtime                  Adds article estimated read time calculator to the site, in the form of '<n> minutes'. 
+Readtime                  Adds article estimated read time calculator to the site, in the form of '<n> minutes'.
 
 Related posts             Adds the ``related_posts`` variable to the article's context
 


nebulousdog:
ok, looks like I was mistaken on the egg fragment thing. see https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support. 

however, we dont have to worry about this if we do the Travis-CI github release + pypi release steps. 

* https://docs.travis-ci.com/user/deployment/pypi/
* https://docs.travis-ci.com/user/deployment/releases/

setup.py
```diff
@@ -0,0 +1,20 @@
+"""minimum config to enable distribution via pip and pipenv."""
+
+import setuptools
+
+with open('README.rst', 'r') as fh:
+    LONG_DESCRIPTION = fh.read()
+
+setuptools.setup(
+    name='pelican-plugins',
+    version='0.0.1',
```
Lucas-C: This should match the latest tag on this repo

Lucas-C:
Two questions:
- with the instructions you added in `Readme.rst`, will it be possible for our users to update `pelican-plugins`, once installed, to the latest commit on `master` ?
- how will the submodules behave ? I do not think they will be initialized with this method. Given the currently high number of submodules in this repo, we may not want to provide a half-functional version of `pelican-plugins` installable with `pip`. See also https://github.com/getpelican/pelican-plugins/issues/1189

nebulousdog:
hey @Lucas-C, thanks for the feedback!

> with the instructions you added in Readme.rst, will it be possible for our users to update pelican-plugins, once installed, to the latest commit on master ?

the short answer is yes! here's one possible approach

1. make a PR as usual with review and Travis-CI kicking off tests as usual. 
2. When merged, a Travis-CI process kicks off adding tags. 
    * [Travis-CI: Setting the tag at deployment time](https://docs.travis-ci.com/user/deployment/releases/#setting-the-tag-at-deployment-time)
3. For your question, when tagged, users now have the most up-to-date reference to the latest commit listed in the repo's [tags page](https://github.com/getpelican/pelican-plugins/tags). 
4. [Travis-CI: PyPI deployment - Deploying tags](https://docs.travis-ci.com/user/deployment/pypi/#deploying-tags)

Lets say we're not ready to deploy to PyPI yet, that's ok, the last step-4 is optional and should probably happen separately afterwards anyway.

Here's the instruction snippet from the [`pipenv`](https://github.com/pypa/pipenv) readme's "Installing from git" section. They didnt make proper headers, so I'll include the part with the `branch_or_tag` bit.

> You can install packages with pipenv from git and other version control systems using URLs formatted according to the following rule:
> 
>     <vcs_type>+<scheme>://<location>/<user_or_organization>/<repository>@<branch_or_tag>#<package_name>

I'm suggesting directly referencing this github repo to incrementally get the installation process started using `pip`, because the PyPI stuff requires extra `twine` registration and adding the pypi encrypted keys as a secret to the `.travis.yml` can happen later.

for your second question

> how will the submodules behave ? I do not think they will be initialized with this method. Given the currently high number of submodules in this repo, we may not want to provide a half-functional version of pelican-plugins installable with pip. See also #1189

I honestly dont know the full implications of mixing submodules with a `pip` install, however it looks supported since https://github.com/pypa/pip/pull/577 went in!

re #1189: sweet, will follow the progress on that.

setup.py
nebulousdog: ok super, I'll add something to keep that in sync. :+1:

Lucas-C:
Thanks for your detailed answer. I fear that I was maybe unclear when I wrote:

> will it be possible for our users to update pelican-plugins

I thought about something like `pip install --upgrade pelican-plugins`. I guess that would only work if frequent package releases of this repo are made, eventually automated with Travis.

nebulousdog:
yeah it can totally be done, but not rushed :D I agree with @moorepants (comment](https://github.com/getpelican/pelican-plugins/issues/425#issuecomment-537103850) and I'm going to close this for now.

Lucas-C:
Ok, great

nebulousdog:
@Lucas-C one question though for going forward, should there be some instructions for adding pelican-plugins in a more modular format, or :scream: is this even the death of the pelican-plugins repo if the recommendation is to package things separately? how do you all want to handle making things being future compatible if people are making new plugins?

Lucas-C:
As explained in other issues, the current approach of the Pelican community toward plugins packaging is to encourage to host them on Pypi. This doesn't mean howerver that this repository will cease to exist, or to welcome new plugins, anytime soon.
@justinmayer : correct me if I'm wrong :)

nebulousdog:
sorry poor phrasing, meant it to mostly be in jest :)

PR #1217, opened 2019-10-01
Simplify Webring plugin HTTPS support

davidag:
After reading that [feedparser will remove its internal HTTP client in following releases](https://github.com/kurtmckee/feedparser/pull/178), it makes sense to stop using it.

This has the added benefit of removing the need to use a global unverified HTTPS context, which could be seen as a security risk.

The _Requests_ library was not used because the requirements were pretty simple and also to avoid another dependency.

### DIFF
--- a/webring/webring.py
+++ b/webring/webring.py
@@ -6,6 +6,8 @@ Webring plugin for Pelican
 A plugin to create a webring in your web site from a list of web feeds.
 """
 import re
+from six.moves.urllib.request import Request, urlopen
+from six.moves.urllib.error import URLError
 from collections import namedtuple
 from logging import warning
 from operator import attrgetter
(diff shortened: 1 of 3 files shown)


webring/webring.py
```diff
@@ -78,35 +80,34 @@ def get_pelican_settings(generators):
...
+    try:
+        req = Request(feed_url)
+        req.add_header(
+            'User-Agent',
+            (
+                'Webring Pelican plugin/{} '
+                + '+https://github.com/pelican/pelican-plugins'
+            ).format(WEBRING_VERSION)
+        )
+        return urlopen(req).read().decode('utf-8')
```
Lucas-C: What would happen in case of a non-200 HTTP status code ?

Lucas-C:
Looks OK to me overall
Thanks :)

webring/webring.py
davidag: Error status codes returned by the server (3xx and 4xx) are handled in the following lines by catching `URLError`. Is that what you refered to?

Lucas-C: Yes. I wasn't sure if an exception would be raised. Thanks for your answer

davidag: Thank you for your review @Lucas-C :tada:

davidag:
@justinmayer May I ask if there is anything else I could do to get this PR merged? Thank you!

justinmayer:
@davidag: Apologies for the delay. Many thanks for the enhancement!

And thanks for reviewing, Lucas. 😁

PR #1222, opened 2019-10-03
Sitemap: Fix configuration example code block rendering

maggick:


maggick:
should fix the sitemap plugin configuration example (github rendering)

justinmayer:
Thank you for the improvement, @maggick.

PR #1223, opened 2019-10-07
Adds Feed Filter plugin

davidag:
Adds a new plugin named **Feed Filter**, which allows to filter elements from feeds according to some user-configured rules.

The use case that inspired me to develop this plugin was to only include articles from some categories in the global feed.

The user can specify what feeds the rules apply to. The rules can be inclusion rules, exclusion rules or both.

Detailed documentation and tests are provided.

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -89,6 +89,8 @@ Events                    Add event start, duration, and location info to post m
 
 Extract table of content  Extracts table of contents (ToC) from ``article.content``
 
+Feed Filter               Filter elements from feeds according to custom rules.
+
 Figure References         Provides a system to number and references figures
 
 Filetime from Git         Uses Git commit to determine page date
@@ -233,7 +235,7 @@ Random article            Generates a html file which redirect to a random artic
 
 Read More link            Inserts an inline "read more" or "continue" link into the last html element of the object summary
 
-Readtime                  Adds article estimated read time calculator to the site, in the form of '<n> minutes'. 
+Readtime                  Adds article estimated read time calculator to the site, in the form of '<n> minutes'.
 
 Related posts             Adds the ``related_posts`` variable to the article's context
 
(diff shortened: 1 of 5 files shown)


avaris:
Multiple `include` or `exclude` patterns behave as `OR` I believe. Might worth testing and mentioning in the readme.

davidag:
@avaris Lot of thanks for your review, I've improved both documentation and tests in order to explicitly mention and test the case of multiple inclusion / exclusion filters.

davidag:
@justinmayer I think it might be interesting to use [the new cookiecutter template](https://github.com/getpelican/cookiecutter-pelican-plugin) for this plugin, but I'm not sure if it's ready yet. What do you think?

PD: When I say ready I mean ready to test it...

justinmayer:
Hey @davidag! I think that's a great idea. I just created the new [pelican-plugins/feed-filter](https://github.com/pelican-plugins/feed-filter/settings/collaboration) repository and invited you to it. The [new cookiecutter template](https://github.com/getpelican/cookiecutter-pelican-plugin) is indeed far enough along to use, so by all means please go for it! It's still a work-in-progress, but we can always re-apply changes as we iterate on the template. Looking forward to seeing your plugin land in the new repository. If you run into any issues, please let us know here or in IRC — I'll be in the [#pelican IRC channel](http://docs.getpelican.com/en/stable/contribute.html#the-pelican-irc-channel) all day today.

davidag:
This plugin is already in the new [pelican-plugins](https://github.com/pelican-plugins) organization and a first version has been [released to pypi](https://pypi.org/project/pelican-feed-filter/) :tada: 

Thanks to @justinmayer for the organization migration and release process, and to @avaris for the code review!

I think this pull request can be closed :relaxed:

justinmayer:
Many thanks for your work in this, David. Glad to see it in its new home.

PR #1224, opened 2019-10-11
Added pelican-gfm plugin

dfoulks1:
There were a few flake8 exceptions in pelican-gfm:
  - __init__.py imports gfm.* but does not call it.
  - a line in gfm.py is too long (left unshortened)

The python unittest ran without error

Created a Readme.md file for this plugin. 10k foot view

Updated the Readme.rst in the pelican-plugins root.

This plugin references https://github.com/getpelican/pelican/issues/2387

### DIFF
--- a/pelican-gfm/Readme.md
+++ b/pelican-gfm/Readme.md
@@ -1,0 +1,22 @@
+pelican-gfm
+===========
+A reader that leverages GitHub's C-based markdown library to translate GitHub Flavored Markdown to html.
+
+Requirements
+============
+
+pelican-gfm has no requirements that are outside of the python standard library aside from pelican itself.
+
+How to Use
+=========
+Drop the entire pelican-gfm directory into the plugin path and invoke it from your pelicanconf.py to have pelican-gfm render contents/\*.md
+
+Syntax
+======
+This plugin leverages [GitHub Flavored Markdown](https://github.github.com/gfm/) in `.md` files to generate html pages.
+
+
+Attribution
+===========
+`pelican-gfm` is based on [pelican_gfm_script](https://github.com/apache/infrastructure-website/blob/master/gfm_reader.py)
+Originally written by: Greg Stein
--- a/pelican-gfm/__init__.py
+++ b/pelican-gfm/__init__.py
@@ -1,0 +1,3 @@
+#!/usr/bin/environment python -B
+
+from gfm import *
--- a/pelican-gfm/gfm.py
+++ b/pelican-gfm/gfm.py
@@ -1,0 +1,208 @@
+#!/usr/bin/python -B
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+#
+# gfm_reader.py -- GitHub-Flavored Markdown reader for Pelican
+#
+
+import sys
+import os
+import ctypes
+import re
+import gfmSetup
+import pelican.utils
+import pelican.signals
+import pelican.readers
+
+_LIBDIR = gfmSetup.LIBCMARKLOCATION
+_LIBCMARK = 'libcmark-gfm.so'
+_LIBEXT = 'libcmark-gfmextensions.so'
+try:
+    cmark = ctypes.CDLL(os.path.join(_LIBDIR, _LIBCMARK))
+    cmark_ext = ctypes.CDLL(os.path.join(_LIBDIR, _LIBEXT))
+except OSError:
+    raise ImportError('%s not found. see build-mark.sh and gfm_reader.py'
+                      % _LIBCMARK)
+
+# Use ctypes to access the functions in libcmark-gfm
+F_cmark_parser_new = cmark.cmark_parser_new
+F_cmark_parser_new.restype = ctypes.c_void_p
+F_cmark_parser_new.argtypes = (ctypes.c_int,)
+
+F_cmark_parser_feed = cmark.cmark_parser_feed
+F_cmark_parser_feed.restype = None
+F_cmark_parser_feed.argtypes = (ctypes.c_void_p,
+                                ctypes.c_char_p,
+                                ctypes.c_size_t,)
+
+F_cmark_parser_finish = cmark.cmark_parser_finish
+F_cmark_parser_finish.restype = ctypes.c_void_p
+F_cmark_parser_finish.argtypes = (ctypes.c_void_p,)
+
+F_cmark_parser_attach_syntax_extension = cmark.cmark_parser_attach_syntax_extension
+F_cmark_parser_attach_syntax_extension.restype = ctypes.c_int
+F_cmark_parser_attach_syntax_extension.argtypes = (ctypes.c_void_p,
+                                                   ctypes.c_void_p,)
+
+F_cmark_parser_get_syntax_extensions = cmark.cmark_parser_get_syntax_extensions
+F_cmark_parser_get_syntax_extensions.restype = ctypes.c_void_p
+F_cmark_parser_get_syntax_extensions.argtypes = (ctypes.c_void_p,)
+
+F_cmark_parser_free = cmark.cmark_parser_free
+F_cmark_parser_free.restype = None
+F_cmark_parser_free.argtypes = (ctypes.c_void_p,)
+
+F_cmark_node_free = cmark.cmark_node_free
+F_cmark_node_free.restype = None
+F_cmark_node_free.argtypes = (ctypes.c_void_p,)
+
+F_cmark_find_syntax_extension = cmark.cmark_find_syntax_extension
+F_cmark_find_syntax_extension.restype = ctypes.c_void_p
+F_cmark_find_syntax_extension.argtypes = (ctypes.c_char_p,)
+
+F_cmark_render_html = cmark.cmark_render_html
+F_cmark_render_html.restype = ctypes.c_char_p
+F_cmark_render_html.argtypes = (ctypes.c_void_p,
+                                ctypes.c_int,
+                                ctypes.c_void_p,)
+
+
+# Set up the libcmark-gfm library and its extensions
+F_register = cmark_ext.core_extensions_ensure_registered
+F_register.restype = None
+F_register.argtypes = ()
+F_register()
+
+# technically, maybe install an atexit() to release the plugins
+
+# Options for the GFM rendering call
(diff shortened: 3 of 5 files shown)


pelican-gfm/gfmSetup.py
```diff
@@ -0,0 +1,111 @@
...
+
+    for package in removed:
+        if dpkg_installed == "0":
+            print(package + " needs removed")
+            return 1
+
+
+def apt_install(package):
+    # I need to be able to do this a better, in a less sudo + apt-y way
+    subprocess.call(["apt-get", "install", package, "-y"])
```
Lucas-C: I don't think those kind of installation steps should be automatically done by plugins.
They'd better be detailed in the plugin README.
What if the those steps require `sudo` rights on the user machine ? What if he uses `apt` or `yum` as a package manager and not `apt-get` ??

pelican-gfm/gfmSetup.py
```diff
@@ -0,0 +1,111 @@
...
+import os
+import subprocess
+
+# This gets used by the gfm plugin as well as the check_configure function
+LIBCMARKLOCATION = "/usr/lib/x86_64-linux-gnu"
+
+ARCHIVES = "https://github.com/github/cmark-gfm/archive"
+VERSION = "0.28.3.gfm.12"
+LOCAL = "cmark-gfm.$VERSION.orig.tar.gz"
+WORKSPACE = '/tmp/build-cmark'
```
Lucas-C: Better use `tempfile.TemporaryDirectiry` : https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

pelican-gfm/Readme.md
```diff
@@ -0,0 +1,22 @@
+pelican-gfm
+===========
+A reader that leverages GitHub's C-based markdown library to translate GitHub Flavored Markdown to html.
+
+Requirements
+============
+
+pelican-gfm has no requirements that are outside of the python standard library aside from pelican itself.
```
Lucas-C: Out of curiosity, have you tried https://pypi.org/project/gfm/ or https://pypi.org/project/py-gfm/ ?
Do they provide what you need maybe, already bundled as Python packages ?

pelican-gfm/Readme.md
```diff
@@ -0,0 +1,22 @@
+pelican-gfm
```
Lucas-C: From my understanding, this plugin will only work on Debian systems.
Better mention this OS limitation in the README

pelican-gfm/gfm.py
```diff
@@ -0,0 +1,208 @@
...
+F_register()
+
+# technically, maybe install an atexit() to release the plugins
+
+# Options for the GFM rendering call
+# this could be moved into SETTINGS or somesuch, but meh. not needed now.
+OPTS = 0
+
+# The GFM extensions that we want to use
+EXTENSIONS = (
```
Lucas-C: Could this be made configurable as a setting ?

pelican-gfm/gfmSetup.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/python -B
+
+import os
+import subprocess
+
+# This gets used by the gfm plugin as well as the check_configure function
+LIBCMARKLOCATION = "/usr/lib/x86_64-linux-gnu"
+
+ARCHIVES = "https://github.com/github/cmark-gfm/archive"
+VERSION = "0.28.3.gfm.12"
```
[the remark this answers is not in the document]
shuttle1987: I'd much prefer this path is configurable because I might not have those packages at `/usr/lib/x86_64-linux-gnu`

pelican-gfm/gfmSetup.py
```diff
@@ -63,82 +62,77 @@ def dpkg_packages_installed():
...
+                                        ".."
+                                      ], cwd=BUILDSPACE)
+            thing1.wait()
+
+            thing2 = subprocess.Popen(["make"], cwd=BUILDSPACE)
+            thing2.wait()
+
+            # Move the libcmark.so artifacts in place
+            print("Moving files")
+            subprocess.call([
```
shuttle1987: Is there some reason you can't use `shutil.move` to move the files here?

pelican-gfm/__init__.py
```diff
@@ -1,3 +1,3 @@
-#!/usr/bin/environment python -B
+#!/usr/bin/environment python333 -B
```
shuttle1987: Is the `python333` a typo?

pelican-gfm/gfmSetup.py
dfoulks1: Used tempfile TemporaryDirectory

pelican-gfm/gfm.py
dfoulks1: this is also now a setting

pelican-gfm/__init__.py
dfoulks1: Yes that was a typo.

pelican-gfm/Readme.md
dfoulks1: I believe that there were pieces of the github markup that those packages did not readily provide. I'm not 100% sure as I Inherited the base of this.

pelican-gfm/gfm.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def add_readers(readers):
+    if str(gfmSetup.test_configuration()) == "0":
+        readers.reader_classes['md'] = GFMReader
+    else:
+        # Optionally we can have the gfmSetup script run here
+        # as root and configure the system
+        # Probably just easier to run it independently tho.
+        sys.exit(1)
```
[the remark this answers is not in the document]
dfoulks1: Done.

pelican-gfm/gfmSetup.py
dfoulks1: I agree, this is a remnant of when this script was a part of something else. I had mistakenly left it in. It's since been removed.

pelican-gfm/gfmSetup.py
dfoulks1: good catch! I ended up going with shutil for these moves.

pelican-gfm/gfmSetup.py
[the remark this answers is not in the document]
dfoulks1: I've put these values in the Settings.py file so that we can update them without having to update any actual code.

pelican-gfm/gfmSetup.py
```diff
@@ -0,0 +1,111 @@
...
+    test_setup()
+    # Configure the environment if it's not already configured
+    if not os.path.isdir(WORKSPACE):
+        os.mkdir(WORKSPACE)
+    subprocess.call(["wget",
+                     "--quiet",
+                     ARCHIVES + "/" + VERSION + ".tar.gz", WORKSPACE,
+                     "-P",
+                     WORKSPACE])
+    subprocess.call(['tar',
```
shuttle1987: Could we replace this `subprocess.call` to `tar -xzf` with using the standard library [`tarfile`](https://docs.python.org/3/library/tarfile.html)?

pelican-gfm/gfmSetup.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def setup():
+    test_setup()
+    # Configure the environment if it's not already configured
+    if not os.path.isdir(WORKSPACE):
+        os.mkdir(WORKSPACE)
+    subprocess.call(["wget",
+                     "--quiet",
+                     ARCHIVES + "/" + VERSION + ".tar.gz", WORKSPACE,
```
shuttle1987: I'd prefer if these paths were created with `os.path.join` such that we can be more easily able to make this cross platform in the future.

pelican-gfm/gfmSetup.py
```diff
@@ -0,0 +1,111 @@
...
+                     "-P",
+                     WORKSPACE])
+    subprocess.call(['tar',
+                     'zxf',
+                     WORKSPACE + "/" + VERSION + ".tar.gz",
+                     "-C",
+                     WORKSPACE])
+    BUILDSPACE = WORKSPACE + "/" + "cmark-gfm-" + VERSION + "/build"
```
shuttle1987: Likewise `os.path.join` would be good for creating the path here

pelican-gfm/gfmSetup.py
dfoulks1: the patsh are now made with os.join

pelican-gfm/gfmSetup.py
dfoulks1: and os.path.join is used here also

shuttle1987: thanks for making these changes

pelican-gfm/gfm.py
```diff
@@ -29,8 +29,16 @@
 import pelican.utils
 import pelican.signals
 import pelican.readers
-from . import gfmSetup
-from . import Settings
+
+try:
```
shuttle1987: What is the purpose of this block with the relative import? If you are trying to maintain some compatibility with ancient Python versions can you solve this problem with the [PEP 328](https://www.python.org/dev/peps/pep-0328/) `from __future__ import absolute_import` approach?

pelican-gfm/gfm.py
```diff
@@ -29,8 +29,16 @@
...
+    from . import gfmSetup
+except ImportError:
+    import gfmSetup
+
+try:
+    from . import Settings
+except:
```
shuttle1987: If you want to use this approach for imports I think this should only catch an `ImportError`.
See my above comment about other potential approaches.

pelican-gfm/test.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/python -B
+
+import unittest
+
+# python2 and python3 differ on how to do this it seems
```
shuttle1987: See my other comment about `from __future__ import absolute_import`

pelican-gfm/gfm.py
```diff
@@ -30,8 +29,8 @@
 import pelican.signals
 import pelican.readers
 
-from . import gfmSetup
-from . import Settings
+import gfmSetup
+import Settings
```
avaris: Can you show *what*'s not working with:

```
from . import gfmSetup
```

That should work.

pelican-gfm/gfmSetup.py
```diff
@@ -0,0 +1,130 @@
...
+
+def dpkg_installed(package):
+    """ Uses Dpkg to determine whether or not a package is installed
+    requires: <package name>"""
+    t1 = subprocess.Popen(["dpkg", "-l"], stdout=subprocess.PIPE)
+    t2 = subprocess.Popen([
+                            "grep",
+                            "-q",
+                            package
+                          ], stdout=subprocess.PIPE, stdin=t1.stdout)
```
avaris: So... you're running `grep` to check for a string in a bunch of lines *from `python`*? :)

pelican-gfm/gfm.py
dfoulks1: I got the absolute imports working but now the plugin doesn't conform to your documentation.

pelican-gfm/gfm.py
dfoulks1: I ended up removing this entirely.

pelican-gfm/gfm.py
dfoulks1: I'm sorry, I was making notes for myself there. I couldn't get the unittest to work and still have test content be generated. that seems to be corrected now.

pelican-gfm/gfmSetup.py
dfoulks1: yeah, that does seem silly. thanks for pointing that out.

PR #1225, opened 2019-10-18
Fixing i18n_subsites/test_i18n_subsites.py

Lucas-C:
Note to reviewers: this test is a pain to maintain, I suggest as an alternative to simply remove it.

What do you think ?


Lucas-C:
@avaris : this should fix the CI
What do you think of the PR & suggestion above ? :)

avaris:
When you say "pain to maintain", is it because of the regeneration? We have similar tests in `pelican` too, and it is not very common that HTML output changes.

Lucas-C:
Yes, I meant that because this test is validating the whole HTML output,
it is not really a **unit** test. It's closer to an end-to-end test, and even there I prefer selectively validating parts of HTML output: check that a given element with an ID or class exist, that it has a given number of children, etc.
Testing the whole HTML output, and requiring it to be identical with even the exact newlines, make those tests very fragile.

avaris:
Yes, it is a full generation test, as said in the class name `TestFullRun` :). It is needed because `i18n_subsites` alters the `pelican` in a somewhat radical way and the output structure will be quite different. This test is a way of validating that. "When run with these inputs, it should generate these files with this folder tree". So, it's not really about matching the HTML inside the files one-to-one, but that's a side effect of using `git diff` for this comparison. Honestly, I think this is the most maintainable way of achieving that, but I'm open to suggestions for alternatives :).

PR #1226, opened 2019-10-19
Add css-html-js-minify plugin

m-col:
Hello,

This PR adds another plugin, pelican-cssmin, as a submodule.

It is essentially [this](https://github.com/zacharyvoase/cssmin), which is a Python-only port of the YUI css compressor. The key differences between the pre-existing yuicompressor plugin is that this version doesn't have the original compressor as a dependency and is fully contained with the relatively small plugin.

I figured I should add this to the pile in case others wanted to use it!

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -73,6 +73,8 @@ Collate content           Makes categories of content available to the template 
 
 Creole reader             Allows you to write your posts using the wikicreole syntax
 
+CSS HTML JS Minify        Minifies all CSS, HTML and JavaScript files in the output path after site generation.
+
 Custom article URLs       Adds support for defining different default URLs for different categories
 
 CTags generator           Generates a "tags" file following the CTags in the "content/" directory, to provide autocompletion for code editors that support it.
--- a/css-html-js-minify/__init__.py
+++ b/css-html-js-minify/__init__.py
@@ -1,0 +1,2 @@
+# -*- coding: utf-8 -*-
+from .css_html_js_minify import *
--- a/css-html-js-minify/css_html_js_minify.py
+++ b/css-html-js-minify/css_html_js_minify.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+"""
+css-html-js-minify wrapper for Pelican
+"""
+
+import glob
+import os
+import sys
+
+from css_html_js_minify import (
+    process_single_css_file,
+    process_single_html_file,
+    process_single_js_file,
+)
+from pelican import signals
+
+def main(pelican):
+    for f in glob.iglob(pelican.output_path + '/**/*.htm*', recursive=True):
+        process_single_html_file(f, overwrite=True)
+    for f in glob.iglob(pelican.output_path + '/**/*.css', recursive=True):
+        process_single_css_file(f, overwrite=True)
+    for f in glob.iglob(pelican.output_path + '/**/*.js', recursive=True):
+        process_single_js_file(f, overwrite=True)
+
+def register():
+    signals.finalized.connect(main)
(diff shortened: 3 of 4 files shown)


Lucas-C:
Thanks for contributing !

FYI, we now also encourage plugin authors to publish their plugins on Pypi.
We are in the progress of polishing & documenting the process. The relavant issue is here : https://github.com/getpelican/pelican-plugins/issues/425 If you go this way, there is already a Pypi classifier and a wiki you can add you plugin to, so it will be referenced: https://github.com/getpelican/pelican-plugins/issues/39#issuecomment-296396499

It is still possible to contribute directly to this repo, but we would prefer directly submitting the code here, in sub-directory, instead of a git submodule, which is burdensome to maintain (less direct control on it, makes setting up this repo a bit more complex & fragile...)
@justinmayer @avaris Do you agree with me ont this ? :)

Finally, we try to encourage plugins to have a few dedicated unit tests, to help detect potential regressions in the future. There are many examples in this repo of such tests, to take inspiration from. Would you please agree to add some tests to you plugin ? :)

justinmayer:
Well, `cssmin` is written in Python, while `yuicompressor` depends on having a Java Runtime Environment (JRE) installed, so that's a very significant difference. That said, [`cssmin`](https://github.com/zacharyvoase/cssmin) isn't maintained any longer, so neither choice is particularly appealing, sadly.

@m-col: Out of curiosity, what was it about `cssmin` that appealed to you? Why choose that tool instead of others? (e.g., [`csscompressor`](https://github.com/sprymix/csscompressor), [`css-html-js-minify`](https://github.com/juancarlospaco/css-html-js-minify#css-html-js-minify), etc.)

m-col:
@justinmayer I initially tried to use cssmin because the code is much smaller than [csscompressor](https://github.com/sprymix/csscompressor) and they compress much the same. I overlooked compressing JavaScript because I don't use any. At the time I started using cssmin it seems I overlooked compressing html as well. In any case, [css-html-js-minify](https://github.com/juancarlospaco/css-html-js-minify) looks superior and would apply too all three filetypes. I would be happy to test it out and report back if there was interest in having it in pelican plugin form, instead of cssmin.

@Lucas-C I am happy to also publish the end result of this plugin to Pypi, and to convert it to direct code instead of a submodule. I've never written any unit tests, but I've been meaning to learn and so I'll check out the other plugins and elsewhere to try to get some set up.

What might be the best way to incorporate a project like [css-html-js-minify](https://github.com/juancarlospaco/css-html-js-minify) as a pelican plugin in such a way that allows future updates from upstream? I could copy over the source folder as-is as into the plugin's folder, so it can be imported by the plugin's own code and used that way, though this would require manual updating by comparing to upstream, without any git tracking. What would be preferable?

m-col:
I updated the plugin to use [css-html-js-minify](https://github.com/juancarlospaco/css-html-js-minify) which is superior to cssmin. This trims the new code down to a small wrapper around css-html-js-minify, which needs to be installed separately.

I read through [#425](https://github.com/getpelican/pelican-plugins/issues/425) to try to understand the new plugin submission process but I couldn't tell if it was in place yet. If it is preferable I can create a standalone repo for this plugin, if that is the desired plugin organisation going forward. If not, the changes I made work very well so the plugin may be complete. Do you have any further suggestions?

justinmayer:
Many thanks for your efforts and patience, @m-col. I think the best path forward is to merge your contribution in its current form. When it comes time to migrate plugins over to the new organization, that will be a good time to make some changes to your plugin so it conforms to the new format and best practices. (See also: https://github.com/getpelican/cookiecutter-pelican-plugin)

Once again, thank you for the contribution!

PR #1229, opened 2019-11-02
md_inline_extension: Fix for Python 3 compatibility

srininara:
substitute basestring with str to work with python3 venv


justinmayer:
Thanks to @srininara for the fix and to @Lucas-C for reviewing.

PR #1240, opened 2020-01-11
Add regex to better guess if there is a math class

geusebi:
When using simple_footnotes plugin the HTML is rewritten omitting quotes where possible.
This cause the test `'class="math"' in content._content` to fail.

This pull request use a regex to make a better guess whether to include the Mathjax script or not.

It should work with single quotes, double quotes, no quotes and multiple classes.
Not a perfect approach but it should cover most cases.

### DIFF
--- a/render_math/math.py
+++ b/render_math/math.py
@@ -32,6 +32,7 @@ the math.  See README for more details.
 
 import os
 import sys
+import re
 
 from pelican import signals, generators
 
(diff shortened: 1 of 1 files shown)


kgaughan:
Hi! I believe that my fix in #1271 should also fix your issue here. I don't suppose you could apply it can see if it works for you?

PR #1243, opened 2020-02-02
fix(liquid_tags.include_code): do not put code inside figcaption. Add CSS classes

talha131:
A regression was introduced few patches ago, which placed code inside
the `figcaption` tag. This is not correct.

The original version has this layout

```
figure
  figcaption
  code
```

This patch fixes this issue. It also encloses metadata,

1. title
2. filename
3. line number

inside CSS classes. This let's themes improve the look of the
figcaption. Previously, all the metadata was piled into one `<span>`.

### DIFF
--- a/liquid_tags/include_code.py
+++ b/liquid_tags/include_code.py
@@ -111,24 +111,29 @@ def include_code(preprocessor, tag, markup):
         raise ValueError("Either title must be specified or filename must "
                          "be available")
 
-    open_tag = ''
-    close_tag = ''
-
-    open_tag = "<figure class='code'>\n<figcaption>"
+    open_tag = "<figure class='code'>\n"
     close_tag = "</figure>"
-    if not hide_all:
-        if not hide_filename:
-            title += " %s" % os.path.basename(src)
-            if lines:
-                title += " [Lines %s]" % lines
-            title = title.strip()
 
-            open_tag += "<span>{title}</span> ".format(title=title)
+    if not hide_all:
+        open_tag+= "<figcaption>"
+
+        if title:
+            open_tag += "<span class=\"liquid-tags-code-title\">{title}</span>".format(title=title.strip())
+
+        if not hide_filename:
+            filename = "%s" % os.path.basename(src)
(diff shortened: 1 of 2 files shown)


talha131:
All the test that have failed in travis, have nothing to do with my patch.

talha131:
Addition of `span` and CSS classes is important. With it themes, can customize the look of metadata. Here is an example.



https://next.elegant.oncrashreboot.com/code-snippets-include-code-from-file

It is not possible if all the metadata is crammed in one span element.

liquid_tags/include_code.py
```diff
@@ -114,22 +114,30 @@ def include_code(preprocessor, tag, markup):
     open_tag = ''
     close_tag = ''
```
avaris: these look redundant

liquid_tags/include_code.py
```diff
@@ -114,22 +114,30 @@ def include_code(preprocessor, tag, markup):
...
+        if lines:
+            lines = " [Lines %s]" % lines
+            open_tag += "<span class=\"liquid-tags-code-lines\">{lines}</span>".format(lines=lines.strip())
 
         if not hide_link:
             url = '/{0}/{1}'.format(code_dir, src)
             url = re.sub('/+', '/', url)
             open_tag += "<a href='{url}'>download</a>".format(url=url)
 
+        open_tag += "</figcaption>"
```
avaris: If you don't have any of these, you still end up with an empty `<figcaption>`. Is that the expected behavior?

talha131: > ff you don't have any of these, you still end up with an empty <figcaption>

No. `<figcaption>` will not be present at all. You can view the source of `pre` [here](https://elegant.oncrashreboot.com/code-snippets-include-code-from-file#example-5-without-metadata) to confirm.

Line 110, which reads

```python
    if (not title and hide_filename) and not hide_all:
        raise ValueError("Either title must be specified or filename must "
                         "be available")
```

makes sure if the code is in this block, then a caption is definitely present.

justinmayer:
Many thanks to @talha131 for the fix and to @avaris for reviewing. 🥇

PR #1258, opened 2020-04-24
Added 'music' plugin: a simple audio file collection presenter.

be1:
For musicians, composers, field recorders, any people having self made audio files, this pelican plugin scans your audio folder, organized by topics and sorted by filename, and generates Album and Track views (if aware ninja2 templates) by exposing article.music_album and article.music_track objects if the article has Album:... or Track:... headers referring to the correct directory or file.

Demo with modified simple HTML template:
http://noise.rocks/~ben/journal/premiere-musique.html
http://noise.rocks/~ben/journal/premier-album.html


be1:
TagPy dependency fails to buld from source with this Travis configuration, missing taglib.h and taglib shared object. I don't know how to fix the build process.

PR #1259, opened 2020-04-30
Asciidoc: Fix "NameError: global name 'tempfile' is not defined"

das-g:
https://github.com/getpelican/pelican/issues/2694#issuecomment-621894352

> That said, I reviewed this thread and made what I surmised to be the relevant changes to the plugin via [getpelican/pelican-plugins@22831d1](https://github.com/getpelican/pelican-plugins/commit/22831d1da399ef54d732c395b0c6a28767622542), so hopefully that fixes the reported problem. If it doesn't, or I screwed it up and broke something horribly, [...]

Well, it does cause `NameError: name 'tempfile' is not defined`. :thinking: 

> [...] by all means please submit a PR in the repository where the plugins actually live. #YOLO

Here you go, @justinmayer :smiley:

### DIFF
--- a/asciidoc_reader/asciidoc_reader.py
+++ b/asciidoc_reader/asciidoc_reader.py
@@ -13,6 +13,7 @@ import os
 import re
 import subprocess
 import sys
+import tempfile
 
 def call(cmd):
     """Calls a CLI command and returns the stdout as string."""


justinmayer:
Nice catch, Raphael. I was almost positive I'd break something, and you proved me right. Bravo, sir. 👍

das-g:
> I was almost positive I'd break something, and you proved me right.

Well, my [Gebastel in the openschoolmaps.ch build](https://gitlab.com/openschoolmaps/openschoolmaps.gitlab.io/-/blob/79e9de3b1c8e8a5b25758c495e45cc19d731bdd7/.gitlab-ci.yml#L81) at [did](https://gitlab.com/openschoolmaps/openschoolmaps.gitlab.io/-/jobs/533884247#L106) and broke https://openschoolmaps.ch.

Guess I shouldn't let it blindly pull from `master` of the repos of others. :see_no_evil:

justinmayer:
Very sorry I caused some disruption. Mea culpa! 🙇

das-g:
Eh, without pinning versions / revisions I kinda had something like that coming.

PR #114, opened 2013-11-29
Added automatic figure numbering to Better Figures & Images Plugin

dflock:
This adds Automatic Figure Numbering to the Better Figures & Images plugin - see [here](http://duncanlock.net/blog/2013/10/19/the-better-figures-images-pelican-plugin-now-supports-figure-numbering/) for more info.

I rebased this as per [this](https://github.com/getpelican/pelican/wiki/Git-Tips#squashing-commits) but it seems to have rebased this onto a commit from Justin? I don't know - I hate git rebase.

Anyway, the diff and commit message both seem to be correct, but since I rebased, github says it's can't auto-merge and I got conflicts (only with my own changes) during the rebase (which I resolved). Sigh. If this won't merge cleanly and you need me to do stuff, just let me know.

PR #1263, opened 2020-05-16
Adds figure numbers to better_figures_and_images

t0rrant:
PR #114 has minor differences that make patching from that commit
impossible. This commit fixed those differences.

Reordered required pip packages to an alphabetical order and added one
missing package, in README.rst

Fixes #114

Signed-off-by: Manuel Torrinha <manuel.torrinha@tecnico.ulisboa.pt>

### DIFF
--- a/better_figures_and_images/README.rst
+++ b/better_figures_and_images/README.rst
@@ -1,7 +1,7 @@
 Requirements
 ------------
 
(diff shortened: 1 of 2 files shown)


t0rrant:
I am now dropping the use of this plugin, and will propose changes to `figure-ref` as it seems more suitable.

PR #1265, opened 2020-06-05
optimize_images: Run jpegtran with -progressive switch

kernc:
Results in progressively-loaded JPEGs and smaller output file size.


justinmayer:
Thanks for the enhancement!

PR #1266, opened 2020-06-05
Plugins shouldn't use root logger

kernc:
See: https://github.com/getpelican/pelican/pull/2763


justinmayer:
Many thanks to @kernc for the enhancements and to @avaris for reviewing. ✨

PR #1270, opened 2020-06-13
fixes for liquid tags

m000:
A bunch of fixes for liquid tags:
  - Tests were horribly broken for python2.7. Made several fixes, now works with `py27-ipython3` and `py37-ipython3`. Tests for the notebook tag still fail.
  - SITEURL is prepended to absolute src paths for the img tag. This feature should probably be ported to other media tags.

Plus a new feature:
  - Support for customizing the Liquid Tag block delimiters. This allows mixing with plugins that also use `{%` and `%}` to mark their blocks (e.g. jinja2content).


kdeldycke:
The `liquid_tags` plugin has been migrated as a stand-alone plugin at https://github.com/pelican-plugins/liquid-tags . All unittests there have been fixed.

But I guess we need @mirekdlugosz to chime in here and re-evaluate all these changes. @mirekdlugosz, what should we do with this PR?

mirekdlugosz:
Most of these changes are valuable and I would welcome ported PR in https://github.com/pelican-plugins/liquid-tags 

There are only two exceptions:
* namespaced plugins were introduced in Pelican 4.5.0, and this is also version that dropped support for Python 2.7, so I'm afraid anything related to Python 2 compatibility should be dropped
* fixes to include_code tag I have applied independently while porting plugin, so they are no longer needed

@m000 is this something you are still interested in? Could you re-apply your changes on top of https://github.com/pelican-plugins/liquid-tags and submit PR there?

m000:
Yes, I plan to migrate the PR. I'll try to find some time for it this week.

PR #1271, opened 2020-06-14
Have simple_bookmarks always quote attributes

kgaughan:
The regex in `Content._get_intrasite_link_regex()` expects attributes to be quoted. When `quote_attr_values` has the value `legacy` when serialising the parsed HTML with html5lib, it ends up omitting the quotes expected, which breaks intrasite links. Changing this to `always` ensures that the quotes remain present if simple_bookmarks processes the page.

I expect this will also fix #1240.

### DIFF
--- a/simple_footnotes/simple_footnotes.py
+++ b/simple_footnotes/simple_footnotes.py
@@ -82,7 +82,7 @@ def parse_for_footnotes(article_or_page_generator):
                     ol.appendChild(li)
                     e.parentNode.removeChild(e)
                 dom.getElementsByTagName(u"body")[0].appendChild(ol)
-                s = html5lib.serializer.HTMLSerializer(omit_optional_tags=False, quote_attr_values='legacy')
+                s = html5lib.serializer.HTMLSerializer(omit_optional_tags=False, quote_attr_values='always')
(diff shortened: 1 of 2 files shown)


justinmayer:
Nice catch, Keith! Thank you for the fix. Since we have been slowly migrating plugins from this monolithic repository to their own individual repositories under the new [Pelican Plugins organization](https://github.com/pelican-plugins), it would be great to also apply these changes to the new [Simple Footnotes plugin repository](https://github.com/pelican-plugins/simple-footnotes). Would it be too much to ask of you to submit a pull request at that repository as well?

Assuming you don't mind, please also include a `RELEASE.md` file with a release type of `patch`, as briefly noted in our [Contributing](https://docs.getpelican.com/en/latest/contribute.html#using-git-and-github) docs.

Please accept my apologies for the confusion. Putting a deprecation notice into the README of migrated plugins is one of the outstanding items on my to-do list, which I think I now have ample reason to prioritize and handle right away. 😅

kgaughan:
Sure! I'll prep another PR there.

kgaughan:
I've created pelican-plugins/simple-footnotes#2, so I'm closing this.

PR #1274, opened 2020-06-21
more_categories: silence printing the category slug

kgaughan:
This gets rid of some noise for anybody using the _more_categories_ plugin.


justinmayer:
Thank you for the fix. We have been slowly migrating plugins from this monolithic repository to their own individual repositories under the new [Pelican Plugins organization](https://github.com/pelican-plugins). As you can see, the new [More Categories plugin repository](https://github.com/pelican-plugins/more-categories) code does not have the `print` statement you removed here.

Please accept my apologies for the confusion. Putting a deprecation notice into the README of migrated plugins is one of the outstanding items on my to-do list. 😅

PR #1276, opened 2020-06-22
[linker] Remove warning for static links

bfueldner:
Add `static` to list of builtin commands and add some explanation to readme file.


justinmayer:
Thanks to @bfueldner for the enhancement and to @avaris for reviewing.

PR #1279, opened 2020-07-07
Replace exec by setattr in neighbors

SG-phimeca:


justinmayer:
Many thanks to @SG-phimeca for the enhancement and to @avaris for reviewing! 💫

Issue #1253, opened 2020-03-30
CI is failing on master


iranzo:
CI should be fixed or master fixed so that it doesn't report as failing and could mask other pr's

PR #1284, opened 2020-07-19
Fix CI tests + update submodules

kernc:
This updates submodules (did check repos for ok-ness/tags/stability) and fixes CI tests. [Except for _summary_ plugin](https://travis-ci.org/github/getpelican/pelican-plugins/builds/709729288#L1204), which was broken with https://github.com/getpelican/pelican/pull/2711 and I don't know what to do about it.

Refs: https://github.com/getpelican/pelican-plugins/issues/1253
Closes: https://github.com/getpelican/pelican-plugins/pull/1225


justinmayer:
Many thanks for these fixes, @kernc.

@galaxy4public: When https://github.com/getpelican/pelican/pull/2711 was submitted, I didn't realize it caused three related tests to fail when merged. Could you possibly have a look and see if you could fix those three failing tests?

avaris:
Ah, this is probably due to the conflicting config variable: `SUMMARY_END_MARKER` added by that PR was used in summary plugin for another purpose. I suppose we can rename the new `pelican` config, given that it is not in a release yet and `summary` plugin was using it for a long time. How about `SUMMARY_END_SUFFIX` instead?

galaxy4public:
> given that it is not in a release yet and `summary` plugin was using it for a long time. How about `SUMMARY_END_SUFFIX` instead?

It is your call whether to fix in in the core or in the plugin.  However, looking into the future, adjusting the core to play nice with the existing plugins has a side-effect of making changes to the core much more difficult (as it was discovered here), since it would mean that the core repository is not self-sufficient to get a confirmation that a change is a non-breaking one.

Please tell me if I need to submit a PR to the core or if we should fix the plugin.  My preference it to make the change to the plugin since a "plugin" is supposed to be a self contained unit.

avaris:
Generally, I agree. I'd change plugin when there is a discrepancy between core and plugin, but I'm leaning towards an exception here:

1. As I said, this is a relatively new addition to core which is not released yet. Very minimal, if any, overhead for renaming. If this was already released, I'd not suggest it.
2. This simply breaks the plugin, and not in a good "throw errors and die" way. It silently "works" but does the wrong thing. Upgraded pelican but not upgraded plugin will be a broken combination that is not obviously apparent. And there is no recourse for the plugin to gracefully move to another config name. 

And no, this does not mean changing core is more difficult. We made a lot of drastic changes to the core in the past that ended up breaking plugins. Plugins (that usually means us, i.e. contributors) had to deal with them and fix it. That's still the case. My suggestion here, in my opinion, is the least headache inducing solution while looking at both sides of the equation.

kernc:
I'll add the `_MARKER` part to me makes somewhat more sense in the plugin compared to core with regard to what the setting does (marks the end of the summary in the source vs. affixes to a summary that's already been extracted and processed).

justinmayer:
@kernc: Glad to see things passing here again. Bravo! Should I squash & merge these now?

kernc:
Not as long as you prefer to visiting Travis to determine whether PRs introduce new breakages. :stuck_out_tongue:

I haven't re-checked updated repos since last month, but I assume their maintainers will do their own due marketing diligence, if that's what you mean. Afaic this is good to go.

justinmayer:
Fantastic. Many thanks for all the time you put into this endeavor! 🥇

PR #1286, opened 2020-08-19
Add note about plugin migration to new organization

### LINKED ISSUE #1280: Migration to https://github.com/pelican-plugins
Hey guys, I found a comment in a recently merged PR that there's an ongoing migration from the monolithic repo to a multi-repo scheme under https://github.com/pelican-plugins.
IMHO this is something that should be prominently displayed in README.md to hint that there may be hiccups in handling PRs and such during the transition period. Perhaps also ask for helping hands if there's anything that users can do to speed up the transition.

justinmayer:
Closes #1280

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -1,13 +1,19 @@
-.. image:: https://travis-ci.org/getpelican/pelican-plugins.svg?branch=master
-    :target: https://travis-ci.org/getpelican/pelican-plugins
-
 Pelican Plugins
 ###############
 
-Beginning with version 3.0, Pelican supports plugins. Plugins are a way to add
-features to Pelican without having to directly modify the Pelican core. Starting
-with 3.2, all plugins (including the ones previously in the core) have been
-moved to this repository, so this is the central place for all plugins.
+**Important note:** We are in the process of migrating plugins from this monolithic repository to their own individual repositories under the new `Pelican Plugins`_ organization, a place for plugin authors to collaborate more broadly with Pelican maintainers and other members of the community. The intention is for all the plugins under the new organization to be in the new “namespace plugin” format, which means these plugins can easily be Pip-installed and recognized immediately by Pelican 4.5+ — without having to explicitly enable them.
+
+This transition process will take some time, so we appreciate your patience in the interim. If you would like to help speed up this transition, the following would be very helpful:
+
+* **If you find a plugin here that has not yet been migrated to the new organization**, create a new issue under this repository and communicate which plugin you would like to help migrate, after which a Pelican maintainer will guide you through the process.
+
+* **If you have come here to submit a pull request to add your plugin**, please consider instead moving your plugin under the `Pelican Plugins`_ organization. To get started, create a new issue under this repository with the details of your plugin, after which a Pelican maintainer will guide you through the process.
+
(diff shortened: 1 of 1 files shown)


Readme.rst
```diff
@@ -1,13 +1,17 @@
...
-with 3.2, all plugins (including the ones previously in the core) have been
-moved to this repository, so this is the central place for all plugins.
+**Important note:** We are in the process of migrating plugins from this monolithic repository to their own individual repositories under the new `Pelican Plugins`_ organization, a place for plugin authors to collaborate more broadly with Pelican maintainers and other members of the community. The intention is for all the plugins under the new organization to be in the new “namespace plugin” format, which means these plugins can easily be Pip-installed and recognized immediately by Pelican 4.5+ — without having to explicitly enable them.
+
+This transition process will take some time, so we appreciate your patience in the interim. If you would like to help speed up this transition, the following would be very helpful:
+
+* **If you find a plugin here that has not yet been migrated to the new organization**, create a new issue under this repository and communicate which plugin you would like to help migrate, after which a Pelican maintainer will guide you through the process.
```
kernc, line 8: A rough up front guideline or a link to a model migrated repo would be nice and maybe result in less guidance.

Readme.rst
```diff
@@ -1,13 +1,17 @@
...
+
+* **If you have come here to submit a pull request to add your plugin**, please consider instead moving your plugin to the `Pelican Plugins`_ organization. Create a new issue under this repository with the details of your plugin, after which a Pelican maintainer will guide you through the process.
```
kernc, line 10: No indication whether the owner then joins the organization and whether they get access rights to their plugin repo.

kernc (approved):

avaris:
Would it be better to open a repo in the pelican-plugins organization for managing administrative requests (and redirect migration related stuff there)? Something like an `_about` repo that would contain a README and would collect issues?

Readme.rst
justinmayer: Good suggestions. I already have a mostly-written plugin migration guide, but it's not quite ready yet. I'll put it up as soon as it's finished, but I don't want that to hold up putting *something* at the top of this README. That said, I just added links to a migrated repo as well as the [Cookiecutter template](https://github.com/getpelican/cookiecutter-pelican-plugin).

Readme.rst
justinmayer: Indeed. The only reason said indication isn't here is that I haven't figured out whether to choose one over the other, or to (continue to) make that call on a case-by-case basis. I'll add that indication as soon as I come up with a better answer. 🤷‍♂️

justinmayer:
@avaris: Perhaps, but better in what way(s)? I'm not saying it isn't better, but it's not immediately clear to me *how* that would be better.

Side note: Why prefix the repo name with an underscore? Does that do something special?

avaris:
Better as in:
* At some point, I expect this repo will be deprecated and in the long run people need a place for requests if they want to add their plugins to the organization structure.
* In the mean time, I suspect there will be some maintenance issues/PRs here. So to not clutter here.

As for the underscore, it was a carry over habit from coding. To differentiate that it is an internal (administrative) repo and not an actual plugin :).

justinmayer:
Thanks for clarifying. I'm actually not sure which repo makes sense to act as the appropriate administrative "desk". This one may work just fine, it may not. But for this moment, it will do the job well enough. We can decide later if it's a good long-term solution.

PR #1288, opened 2020-08-26
Add user-facing GZIP_CACHE_EXCLUDE_TYPES file type exclusion list

### LINKED ISSUE #852: gzip_cache: Potential for a user-facing exclude option?
I recently started using the `gzip_cache` plugin on my site, and I am shipping some extra files through Pelican that do not benefit from compression really, and just clutter up the output folder. Mostly because they will not be used in the classic "serve a static HTML website" kind of way, but more as direct file downloads.

Therefore I contemplated extending the plugin to implement such a feature in compliance with other Pelican `pelicanconf.py` options à la `PAGE_EXCLUDES`, `ARTICLE_EXCLUDES`, and the like. I just wanted to check in before going about it, if there are other opinions about such a feature?

sagamusix:
Closes #852

### DIFF
--- a/gzip_cache/gzip_cache.py
+++ b/gzip_cache/gzip_cache.py
@@ -16,7 +16,7 @@ from pelican import signals
 logger = logging.getLogger(__name__)
 
 # A list of file types to exclude from possible compression
-EXCLUDE_TYPES = [
+EXCLUDE_TYPES = (
     # Compressed types
     '.bz2',
     '.gz',
@@ -43,7 +43,7 @@ EXCLUDE_TYPES = [
     # but it's not worth it.
     '.woff',
     '.woff2',
-]
+)
 
 COMPRESSION_LEVEL = 9 # Best Compression
 
@@ -63,21 +63,25 @@ def create_gzip_cache(pelican):
 
     :param pelican: The Pelican instance
     '''
+    user_exclude_types = pelican.settings.get('GZIP_CACHE_EXCLUDE_TYPES', ())
+    overwrite = should_overwrite(pelican.settings)
     for dirpath, _, filenames in os.walk(pelican.settings['OUTPUT_PATH']):
         for name in filenames:
-            if should_compress(name):
+            if should_compress(name, user_exclude_types):
                 filepath = os.path.join(dirpath, name)
-                create_gzip_file(filepath, should_overwrite(pelican.settings))
+                create_gzip_file(filepath, overwrite)
 
 
-def should_compress(filename):
+def should_compress(filename, user_exclude_types):
     '''Check if the filename is a type of file that should be compressed.
 
     :param filename: A file name to check against
+    :param user_exclude_types: User-supplied list of extensions to exclude
     '''
-    for extension in EXCLUDE_TYPES:
-        if filename.endswith(extension):
-            return False
+    if filename.endswith(EXCLUDE_TYPES):
+        return False
+    if filename.endswith(user_exclude_types):
+        return False
 
     return True
 
--- a/gzip_cache/Readme.rst
+++ b/gzip_cache/Readme.rst
@@ -12,6 +12,10 @@ file within the same directory as the original file.
 Settings
 --------
 
-* `GZIP_CACHE_OVERWRITE`
+* ``GZIP_CACHE_OVERWRITE``:
   If True, the original files will be replaced by the gzip-compressed files. 
   This is useful for static hosting services (e.g S3). Defaults to False.
+* ``GZIP_CACHE_EXCLUDE_TYPES``:
+  A tuple of file extensions to exclude in addition to the built-in ones.
+  Extensions must be specified with leading dot, e.g.
+  ``GZIP_CACHE_EXCLUDE_TYPES = ('.php', '.py', '.tmp')``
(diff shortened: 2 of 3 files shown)


gzip_cache/gzip_cache.py
```diff
@@ -65,19 +65,23 @@ def create_gzip_cache(pelican):
     '''
     for dirpath, _, filenames in os.walk(pelican.settings['OUTPUT_PATH']):
         for name in filenames:
-            if should_compress(name):
+            if should_compress(name, pelican.settings):
```
avaris, line 68: Passing the whole setting to `should_compress` seems excessive when it only cares about a single option. Would be better to extract that setting here and pass that.

Bonus: Minimal performance boost since you'll do that once instead of every file.

gzip_cache/gzip_cache.py
```diff
@@ -65,19 +65,23 @@ def create_gzip_cache(pelican):
...
+def should_compress(filename, settings):
     '''Check if the filename is a type of file that should be compressed.
 
     :param filename: A file name to check against
+    :param settings: The pelican instance settings
     '''
     for extension in EXCLUDE_TYPES:
         if filename.endswith(extension):
             return False
+    for extension in settings.get('GZIP_CACHE_EXCLUDE_TYPES', {}):
```
avaris, line 82: Setting is a `list`, so why is the fallback value a `dict`?

gzip_cache/gzip_cache.py
```diff
@@ -65,19 +65,23 @@ def create_gzip_cache(pelican):
...
+        if filename.endswith(extension):
+            return False
```
avaris, line 84: FWIW, `.endswith` can accept a tuple of values which would eliminate external loop. Not necessary, but felt like mentioning. Similarly for above.

gzip_cache/gzip_cache.py
sagamusix: I was just doing exactly the same way as the other helper function in the same file. Should I adjust both helper functions to work that way?

gzip_cache/gzip_cache.py
sagamusix: good point

avaris:
Also, needs documentation in the README for the new setting.

gzip_cache/gzip_cache.py
sagamusix: I just did it this way because that's how the extension already did it. When in Rome, do as the Romans do ;) Should I modify both loops?

avaris: Like I said, not necessary. Mostly a style preference. Up to you :).

gzip_cache/gzip_cache.py
sagamusix: I think `should_overwrite` was implemented this way to make the settings value testable. If it's evaluated in `create_gzip_cache`, its effect can no longer be observed. Should I remove the test for `should_overwrite` or leave `should_overwrite` untouched?

gzip_cache/gzip_cache.py
sagamusix: Done

gzip_cache/gzip_cache.py
sagamusix: Nevermind, I can keep the function but only read it once instead for each file. That will simplify things at least a bit.

sagamusix: I pushed the simplification as a separate commit.

gzip_cache/gzip_cache.py
sagamusix: I pushed the simplification as a separate commit.

sagamusix:
All requested changes have been applied and the README adjusted.

gzip_cache/gzip_cache.py
avaris: I mean, sure it is testable but also not much point in testing `.get` method of `dict`s :).

gzip_cache/gzip_cache.py
```diff
@@ -77,12 +78,10 @@ def should_compress(filename, user_exclude_types):
...
-    for extension in EXCLUDE_TYPES:
-        if filename.endswith(extension):
-            return False
-    for extension in user_exclude_types:
-        if filename.endswith(extension):
-            return False
+    if filename.endswith(EXCLUDE_TYPES):
+        return False
+    if filename.endswith(user_exclude_types):
+        return False
```
avaris, line 84: `endswith` is picky about type, it really wants `tuple`s.

sagamusix: Just noticed :) It's fixed now.

sagamusix:
Should I maybe convert `EXCLUDE_TYPES` and the new user setting from lists to tuples, so that they don't have to be converted to tuples inside `should_compress` again and again?

avaris (approved):
Thank you! 👍

sagamusix:
Okay, converted to tuples as it wasn't really much extra work.

sagamusix:
Can we get this merged?

justinmayer:
Will have a look as soon as I can...

sagamusix:
Ping? :)

justinmayer:
Sincere apologies for the delay, @sagamusix. Many thanks for your work on this, and thanks to @avaris for reviewing! ✨

Issue #1290, opened 2020-08-28
[related_posts, thumbnailer, tipue_search] Migration out of mono-repo


kdeldycke:
In the same spirit as https://github.com/getpelican/pelican-plugins/issues/1287 , I'll be happy to assist in the migration of the following plugins into their own repositories:

* [x] [`thumbnailer`](https://github.com/getpelican/pelican-plugins/tree/master/thumbnailer)
* [x] [`related_posts`](https://github.com/getpelican/pelican-plugins/tree/master/related_posts)
* [x] [`tipue_search`](https://github.com/getpelican/pelican-plugins/tree/master/tipue_search)

justinmayer:
Fantastic! Why don't we start next with the Thumbnailer plugin, which I just created a repository for here: https://github.com/pelican-plugins/thumbnailer

justinmayer:
@kdeldycke: I had a few minutes, so I decided to [help out with the Related Posts plugin](https://github.com/pelican-plugins/related-posts/pull/1).

justinmayer:
Related Posts is done!  🎉  On to the next one...

Speaking of which, should we still call it Tipue Search? Since everything about the upstream project has been orphaned/abandoned, I'm wondering whether this is a good time to rename it to something else. Any thoughts on that topic?

kdeldycke:
I want to play it safe with Tipue Search: keep maintain it for a while and have a practical solution right now. I'm no certain I'll have enough time or energy to start searching and investing in a brand new solution (like `lunr.js`). I feel like I'm shaving that yak for way too long! 😆

justinmayer:
Totally understandable. I agree! To be clear, I wasn't suggesting starting over from scratch with a new search solution, but instead perhaps changing the plugin *name*, in large part because we may indeed swap out the engine powering the search functionality in the future.

kdeldycke:
@justinmayer OK I get it. But I'm worried about the limited maintainership resources around Pelican and its ecosystem.

Which makes me feel it's going to be easier, as a first step, to keep the Tipue Search plugin in its current state, and only make the minimal investment to have it just work as an independent plugin.

Then we'll be free in the future to mark Tipue Search as deprecated, if a better rock solid solution materialize and gain traction. That way we're not blocking:
* people searching for a working solution right now (Tipue Search plugin)
* people working on a better solution solution (lunr.js-based or other)

The risk is starting making plans about an ideal generic plugin, to finally realize we're not having the manpower to implement them.

[...]

PR #1289, opened 2020-08-28
Add neighbors plugin migration disclaimer.

kdeldycke:
Add deprecation warning in the same spirit as: https://github.com/getpelican/pelican-plugins/commit/f8394e39df71f697c6879ef88e16f4d6d4bce614

Relates to:
* https://github.com/getpelican/pelican-plugins/issues/1287
* https://github.com/pelican-plugins/neighbors/pull/2

PR #1292, opened 2020-08-30
Add disclaimer for migrated plugins.

kdeldycke:
`related_posts` and `thumbnailer` plugins have been moved to their own repository. Add disclaimers in the same spirit as https://github.com/getpelican/pelican-plugins/pull/1289.

Relates to https://github.com/getpelican/pelican-plugins/issues/1290.

### DIFF
--- a/related_posts/Readme.rst
+++ b/related_posts/Readme.rst
@@ -1,6 +1,10 @@
 Related posts
 -------------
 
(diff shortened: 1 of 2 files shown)


justinmayer:
Many thanks for remembering this step, Kevin! 🏅

PR #1296, opened 2020-08-31
[Migration] Flag Tipue Search as migrated

kdeldycke:
`tipue_search` plugin has been moved to its own repository. Add disclaimers in the same spirit as https://github.com/getpelican/pelican-plugins/pull/1289 and https://github.com/getpelican/pelican-plugins/pull/1292.

Relates to https://github.com/getpelican/pelican-plugins/issues/1290.

Depends on https://github.com/pelican-plugins/tipue-search/pull/1.

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -318,7 +318,7 @@ Sitemap                                                           `✔  <https:/
 
 Thumbnailer                                                       `✔  <https://github.com/pelican-plugins/thumbnailer>`_                    Creates thumbnails for all of the images found under a specific directory
 
-`Tipue Search <./tipue_search>`_                                                                                                            Serializes generated HTML to JSON that can be used by jQuery plugin - Tipue Search
+Tipue Search                                                      `✔  <https://github.com/pelican-plugins/tipue-search>`_                   Serializes generated HTML to JSON that can be used by jQuery plugin - Tipue Search
 
 `Touch <./touch>`_                                                                                                                          Does a touch on your generated files using the date metadata from the content
 
--- a/tipue_search/README.md
+++ b/tipue_search/README.md
@@ -1,6 +1,10 @@
 Tipue Search
 ============
 
+**NOTE:** [This plugin has been moved to its own repository](https://github.com/pelican-plugins/tipue-search). Please file any issues/PRs there. Once all plugins have been migrated to the [new Pelican Plugins organization](https://github.com/pelican-plugins), this monolithic repository will be archived.
+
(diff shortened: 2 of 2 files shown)


kdeldycke:
This plugin has been migrated, we can now merge this PR.

justinmayer:
So, the reason I have been holding off on moving forward with publishing the search plugin to PyPI relates to my [previous comment](https://github.com/getpelican/pelican-plugins/issues/1290#issuecomment-683375633) on the topic.

@kdeldycke: [Your response](https://github.com/getpelican/pelican-plugins/issues/1290#issuecomment-683420404) makes sense, but I think it’s predicated on a misunderstanding of what I meant. My only suggestion was that perhaps we could use a different name, leaving everything else the same. As background information, I wrote down a few related thoughts regarding [plugin migration philosophy](https://github.com/getpelican/pelican-plugins/issues/1311#issuecomment-786633254), which might shed more light on why I think this topic matters. Creating a separate repository and corresponding PyPI package for what will (hopefully) be a short-lived, point-in-time snapshot of a deprecated and unsupported plugin… It just doesn’t make a lot of sense to me.

Instead, what I would prefer to do might look something like this:

* make the absolute minimum amount of changes needed to support the project name change
* publish that code as v0.1
* hopefully, eventually publish v1.0 based on a supported foundation

That was the intention behind my original proposal to change the name, which would avoid the previously-raised concerns:

* blocking people searching for a working solution right now (Tipue Search plugin)
* blocking people working on a better solution solution (lunr.js-based or other)
* making plans about an ideal generic plugin, only to realize don’t have the manpower to implement them

To that end, I have pushed a new [`rename` branch](https://github.com/pelican-plugins/tipue-search/tree/rename) with the aforementioned minimal changes. I hope that it will simultaneously address my concerns about dragging around dead repos/packages while still allowing folks to have a working solution right now.

@kdeldycke: Would you have a look at the [new branch](https://github.com/pelican-plugins/tipue-search/tree/rename) and let me know what you think? I tried to use your blog site content and Plumage theme to test the renamed search plugin, but I ran into unrelated dependency conflicts and ultimately had to abort. I understand if you don’t have time to test out the new branch to ensure it works but would of course appreciate it if you don’t mind too much. 😅

kdeldycke:
Sorry @justinmayer for the late reply. Now that I backlogged the cross-comments, I have a better understanding of your proposal. So... let's do it! Let's merge your `rename` branch. Pretty sure it will work as-is. I'll test it later. Don't wait for me ! :)

PR #1299, opened 2020-09-17
[i18n_subsites]: Fix broken feed links

### LINKED ISSUE #1033: i18N_subsites and feed generation
Hello everyone,

I don't seem to be able to find it, is there a pre-thought way of having feeds for different language in different atom files? I don't think readers of one language care about the translations of the same article...

I can work around it by using categories, but it's a bit of an hack.

Thanks!

jorgesumle:
I think it fixes #1033

### DIFF
--- a/i18n_subsites/i18n_subsites.py
+++ b/i18n_subsites/i18n_subsites.py
@@ -104,6 +104,8 @@ def prepare_site_db_and_overrides():
             overrides['THEME_STATIC_DIR'] = posixpath.join(
                 relpath, _MAIN_SETTINGS['THEME_STATIC_DIR'])
             overrides['THEME_STATIC_PATHS'] = []
+        if 'FEED_DOMAIN' not in overrides:
+            overrides['FEED_DOMAIN'] = posixpath.join(main_siteurl, lang)
         # to change what is perceived as translations
         overrides['DEFAULT_LANG'] = lang
 


i18n_subsites/i18n_subsites.py
```diff
@@ -104,6 +104,8 @@ def prepare_site_db_and_overrides():
             overrides['THEME_STATIC_DIR'] = posixpath.join(
                 relpath, _MAIN_SETTINGS['THEME_STATIC_DIR'])
             overrides['THEME_STATIC_PATHS'] = []
+        if 'FEED_DOMAIN' not in overrides:
+            overrides['FEED_DOMAIN'] = posixpath.join(main_siteurl, lang)
```
avaris, line 108: This should probably use `settings['FEED_DOMAIN']` instead of `main_siteurl`

jorgesumle: I changed what you said @avaris, but I think it's not the best idea. If you have `FEED_DOMAIN` set to `SITEURL + '/feeds'`, for example, the link to the feeds for Czech will be like `http://example.com/test/feeds/cz/all.atom.xml`, and this plugin doesn't organize the output that way, so the links are broken.

In this case, when `FEED_DOMAIN` is `None`, it works just fine, but if it's provided it makes things more difficult and unexpected with that change.

Issue #1059, opened 2018-09-12
[tag cloud] Tag cloud emits fewer steps than expected (improvement)


tofof:
### Summary
In many cases where the max_count for any particular tag and the number of steps are of similar magnitude, tag_cloud will emit only n-1 steps even when it would be possible to emit n steps. 

### Expected behavior
Tag_cloud makes use of all n steps.

### Example
min_count 1, max count 5, steps 4 -- assume tags exist for all possible sizes (1,2,3,4,5). Tag_cloud will emit tag-4, tag-2, and tag-1.  No tags will be assigned tag-3, since [the formula](https://github.com/getpelican/pelican-plugins/blob/master/tag_cloud/tag_cloud.py#L54-L55) produces 2.7 for size 2 tags, which gets floored to 2.

size| result | floored | rounded
--- | --- | --- | ---
1 |	4 | 4 | 4
2 |	2.71 | 2 | 3
3 |	1.95 | 1 | 2
4 |	1.42 | 1 | 1
5 |	1 | 1 | 1

### Suggested resolution
If the formula used `round` instead of `math.floor`, this would naturally resolve the issue in most (though not all) cases:

min_count | max_count | steps |	steps used (floor) | steps used (round) | result |
|-----|--------|------------|-----------|------|--------|
 1 |  3  |  3  | 2  | 3  | ✅ improved |
 1 |  3  |  4  | 3  | 3  | 👌 ok  |
 1 |  4  |  4  | 3  | 4  | ✅ improved |
 1 |  5  |  4  | 3  | 4  | ✅ improved |
 1 |  6  |  4  | 3  | 4  | ✅ improved |
 1 |  7  |  4  | 3  | 4  | ✅ improved|
 1 |  8  |  4  | 4  | 4  | 👌 ok  |
 1 |  4  |  5  | 4  | 4  | 👌 ok |
 1 |  5  |  5  | 4  | 4  | ❌ bad in both |
 1 |  6  |  5  | 4  | 4  | ❌ bad in both  |
 1 |  7  |  5  | 4  | 5  | ✅ improved  |
 1 |  8  |  5  | 4  | 5  | ✅ improved |
 1 |  9  |  5  | 4  | 5  | ✅ improved  |
 1 |  10 |   5 |  4 |  5 |  ✅ improved  |
 1 |  11 |   5 |  4 |  5 |  ✅ improved  |
 1 |  12 |   5 |  4 |  5 |  ✅ improved  |
[...]

PR #1302, opened 2020-10-05
[Tag cloud] use round instead of Math.floor

### LINKED ISSUE #1058: [tag cloud] Tag cloud emits tag-0 class (bug)
### Observed behavior
When min_count is 1, max_count is 6, and steps is the default of 4, the emitted class for the tags of size 6 is 0, not 1.

### Expected behavior
Tag classes range from tag-1 to tag-N, as specified in readme:
> You should then also define CSS styles with appropriate classes (tag-1 to tag-N, where N matches TAG_CLOUD_STEPS)

### Suggested resolution
Change [this formula](https://github.com/getpelican/pelican-plugins/blob/master/tag_cloud/tag_cloud.py#L54) to `round` instead of `math.floor`. This actually has [other useful effects](https://github.com/getpelican/pelican-plugins/issues/1059) in addition to correcting what seems to be a floating point error.

jorgesumle:
It fixes #1058 and #1059

### DIFF
--- a/tag_cloud/tag_cloud.py
+++ b/tag_cloud/tag_cloud.py
@@ -51,7 +51,7 @@ def generate_tag_cloud(generator):
     def generate_tag(tag, count):
         tag = (
             tag,
-            int(math.floor(steps - (steps - 1) * math.log(count - min_count + 1)
+            int(round(steps - (steps - 1) * math.log(count - min_count + 1)
                 / (math.log(max_count - min_count + 1) or 1)))
         )
         if generator.settings.get('TAG_CLOUD_BADGE'):


justinmayer:
I'm so sorry, @jorgesumle! [I migrated the Tag Cloud plugin to its own repository](https://github.com/pelican-plugins/tag-cloud) before noticing your open pull request. Would you mind making this change in a new pull request at the plugin's new location? I could make the change myself, of course, but would prefer that you get credit for it. 😊 

Assuming that's okay and you don't mind re-submitting this change, would you also include a [`RELEASE.md`](https://docs.getpelican.com/en/latest/contribute.html#using-git-and-github) file in the PR so a new release is issued when the PR is merged?

justinmayer:
Fix applied and released in Tag Cloud 1.0.1: https://github.com/pelican-plugins/tag-cloud/releases/tag/1.0.1

PR #1227, opened 2019-10-26
Image process pypi warnings

Ecno92:
Image process is now available on Pypi and the commit referenced in this project is outdated. 
By updating the reference to a commit with a warning and by updating the readme I hope to make sure that all users can transition smoothly to the next version. 

Let me know if everything is fine like this or whether adjustments are necessary.

PR #1307, opened 2020-11-15
Flag image-process plugin as migrated

kdeldycke:
`image-process` plugin has just been moved to `pelican-plugins` organization and is about to be release as an independent Python package for the first time. It can be flagged as migrated.

See: https://github.com/pelican-plugins/image-process/issues/32#issuecomment-727580878

This closes #1227.


justinmayer:
Many thanks, Kevin.

PR #1310, opened 2020-12-01
AsciiDoc Reader: Support Unicode Metadata

podsvirov:
With this changes I can render Cyrillic posts with Cyrillic metadata via `asciidoctor` command.

Without this changes I get error:
```
UnicodeEncodeError: 'latin-1' codec can't encode characters
```

And non `ASCII` metadata not parsed.

I use `pelican 4.2.0`.

### DIFF
--- a/asciidoc_reader/asciidoc_reader.py
+++ b/asciidoc_reader/asciidoc_reader.py
@@ -25,14 +25,6 @@ def default():
         if len(call(cmd + " --help")):
             return cmd
 
-def fix_unicode(val):
-    if sys.version_info < (3,0):
-        val = unicode(val.decode("utf-8"))
-    else:
-        # This fixes an issue with character substitutions, e.g. 'ñ' to 'Ã±'.
-        val = str.encode(val, "latin-1").decode("utf-8")
-    return val
-
 ALLOWED_CMDS = ["asciidoc", "asciidoctor"]
 
 ENABLED = None != default()
@@ -74,7 +66,7 @@ class AsciiDocReader(BaseReader):
         """Parses the AsciiDoc file at the given `source_path` and returns found
         metadata."""
         metadata = {}
-        with open(source_path, encoding="latin-1") as fi:
+        with open(source_path, encoding='utf-8') as fi:
             prev = ""
             for line in fi.readlines():
                 # Parse for doc title.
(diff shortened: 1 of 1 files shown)


podsvirov:
Looks like all good for `asciidoc_reader`:
```
Summary:
         top-module | success | error   | skip   
   -------------------------------------------------
    asciidoc_reader |       2 |       0 |       0
```

podsvirov:
Cc: @avaris @nickoe @jichu4n @sabaini @marco-m @jeffrimko @nikolausschueler @justinmayer @das-g

I apologize for the unnecessary noise. I am just trying to use this plugin to create Cyrillic (input files encoding is utf-8) articles using the [pelican](https://packages.msys2.org/base/mingw-w64-pelican) from the [MSYS2](https://www.msys2.org/) distribution.

asciidoc_reader/asciidoc_reader.py
```diff
@@ -74,7 +74,7 @@ def _read_metadata(self, source_path):
         """Parses the AsciiDoc file at the given `source_path` and returns found
         metadata."""
         metadata = {}
-        with open(source_path) as fi:
+        with open(source_path, encoding="latin-1") as fi:
```
das-g, line 77: Wait, `latin-1`? According to the AsciiDoctor FAQ, AsciiDoctor treats input files as UTF-8 by default and as UTF-16 if there's a BOM. Shouldn't we mimic the same behavior here?

das-g (commented):
The default encoding of Python's open is platform specific, so explicitly specifying an encoding is a Good Thing™. But is `latin-1` really the right one?

asciidoc_reader/asciidoc_reader.py
podsvirov: I don't like that either. The root of this problem is in the implementation of the `fix_unicode` function for `python3`.
Another way for me is to remove `val = str.encode(val, "latin-1"). decode("utf-8")` from `fix_unicode` function.

avaris: I agree. Encoding should be `utf-8` and after that as far as I can see, `fix_unicode` can be removed. It shouldn't be necessary.

podsvirov: Okay, done in second commit.

podsvirov:
What about to merge this PR? Any additional conversation or checks?

das-g (approved):
LGTM

avaris (approved):

justinmayer:
Many thanks to @podsvirov for the enhancement and to @das-g and @avaris for reviewing. 💫

PR #1319, opened 2021-01-26
[Migration] Flag Show Source plugin as migrated

kdeldycke:
Depends on https://github.com/pelican-plugins/show-source/issues/1


justinmayer:
Thanks, Kevin! 🥇

PR #1339, opened 2021-05-25
Fix plugin descriptions table

gadiyar:
The plugin description table wasn't showing up. Added one space before the description on every line to fix this.


justinmayer (approved):
Many thanks for fixing this problem. Much appreciated! 🏅

Issue #1342, opened 2021-06-21
[Migration] photos


phibos:
I'm not the original author of the photos plugin, but I would like to migrate it to the new organisation and provide some improvements.

phibos:
I'm currently migrating the repo. Feel free to have a look at https://github.com/phibos/pelican-photos . Hope this helps. Not finished yet, some tests still fail.

phibos:
All issues have been fixed and an admin can push the repository into the new pelican-plugins group.

@justinmayer

phibos:
I have also created a PR to mark the photos plugin as migrated.

@justinmayer

phibos:
Anything I can do to speed up the process? @MinchinWeb @kernc @kdeldycke @davidlesieur

justinmayer:
Hi @phibos. Thank you for the offer to assist with the Photos plugin migration. I have many other high-priority items competing for my attention at this time, but I rest assured I will follow up with you as soon as I can. Have a great day! 😁

paulo-coutinho:
Hi,

I have erros with photos plugin:

```
DEBUG: Pelican version: 4.6.0
DEBUG: Python version: 3.9.5
DEBUG: Adding current directory to system path
DEBUG: Finding namespace plugins
DEBUG: Namespace plugins found:
  | pelican.plugins.webassets
DEBUG: Loading plugin `assets`
DEBUG: Loading plugin `sitemap`
DEBUG: Loading plugin `summary`
DEBUG: Loading plugin `extended_meta`
DEBUG: Loading plugin `related_posts`
DEBUG: Loading plugin `photos`
DEBUG: piexif found.
DEBUG: Registering plugin `assets`
DEBUG: Registering plugin `sitemap`
DEBUG: Registering plugin `summary`
DEBUG: Registering plugin `extended_meta`
DEBUG: Registering plugin `related_posts`
DEBUG: Registering plugin `photos`
  --- AutoReload Mode: Monitoring `content`, `theme` and `settings` for changes. ---

Serving site at: http://127.0.0.1:8000 - Tap CTRL-C to stop

-> Modified: settings, content, theme, [static]images, [static]extras. re-generating...
Process SpawnPoolWorker-1:1:
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
[...]

PR #1343, opened 2021-06-30
Mark photos plugin as migrated

phibos:
After #1342 has been finished this marks the photos plugin as migrated.


justinmayer:
Thank you, @phibos, for all your help with the migration. Version 1.0 of the newly migrated namespace plugin has been published to PyPI.

PR #1345, opened 2021-08-11
[pelican_comment_system] Fix _get_writer() call

codeurimpulsif:
## Goal

This pull request fix the following error with the `pelican_comment_system` plugin:
```
CRITICAL: 'Pelican' object has no attribute 'get_writer'
```

## Explanation

Since https://github.com/getpelican/pelican/pull/2821 `get_writer()` is renamed `_get_writer()`, this pull request only rename this function call.


justinmayer:
Thanks to @codeurimpulsif for the contribution and to @avaris for reviewing. 😁

PR #1354, opened 2021-12-18
[events] fixed: populate events-list template variable with the Article() objects

moormaster:
… instead of just lightweight Event()-objects. This enables the template to place links to the real article url. Pages therefore do not need to have a slug header anymore for these links to work.


justinmayer:
Thank you for the enhancements, @moormaster!

PR #1371, opened 2022-11-15
asciidoc_reader: Don't process content twice

dflock:
We process the content twice, once on line 63, then again on 67 - just keep the last one.


justinmayer:
Thanks, Duncan!

Issue #1379, opened 2023-01-24
An alternate presentation of pelican-plugins and legacy plugins repos


maphew:
It's taken me awhile to wrap my ahead around the pre- and post- Pelican 4.5 plugin repositories and organization. To help others grapple with the Pelican universe I think it would be good make a clearer distinction between the old and new structures. 

To that end I've forked the repository and re-arranged in a way that I think is clearer, while still allowing folks who need to use the old plugins to do so with a minimum change in workflow. I'll reference this issue shorty with a PR that shows what I'm thinking of.

PR #1380, opened 2023-01-24
New default main branch

maphew:
See #1379. PR 1 of 2. 

Changes: 

- rename default branch `master` to `main`
- remove plugin folders
- shorten Readme to only post 4.5, with short intro to Legacy for those that need it


egberts:
What?  It has been co-opted that  plugin folders are here to stay at this stage. 

 IMHO, this pull request should be three separate pull requests.

maphew:
Hi! This PR is a year and half old and is probably no longer relevant. I'd forgotten about it at any rate! I ended up not using Pelican in the project that spawned this, and the details that made it seem like a good idea are not coming to mind. I'll close it.

PR #1393, opened 2023-11-10
`bootstrap-rst`: Math header is an array, so should be joined not appended

robtaylor:
bootstrap-rst was crashing with recent docutils, issue was the handling of math_header which caused an excpetion in `docutils/writers/_html_base.py`:
```
                    ", line 167, in interpolation_dict
                        subs[attr] = ''.join(getattr(self,
                    attr)).rstrip('\n')
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                    TypeError: sequence item 2: expected str instance,
                    list found
```


justinmayer:
Many thanks for this fix, Rob! 🥇

PR #1398, opened 2024-02-15
org_reader: Add support for custom templates

azjf:
@see https://docs.getpelican.com/en/latest/faq.html#how-do-i-assign-custom-templates-on-a-per-page-basis


justinmayer (approved):
Thank you for the enhancement, @azjf 🏅

PR #1399, opened 2024-04-08
Pillow update 10.0.0

fantomfp:
`Image.ANTIALIAS` was [removed in Pillow 10](https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#constants). Instead, you can use `Image.LANCZOS` or `Image.Resampling.LANCZOS`.


justinmayer:
Hi Casper. As noted at the top of the respective plugins' `README` files, these plugins have been migrated to their own individual repositories, and any and all issues/PRs should be filed in those individual repositories.

However, in this case, the changes needed for Pillow 10+ have already been made:

* https://github.com/pelican-plugins/photos/pull/90
* https://github.com/pelican-plugins/thumbnailer/pull/11

Issue #1401, opened 2024-06-05
[asciidoc_reader] Issue with detecting available binaries


tpwo:
Hi, I noticed an issue with `asciidoc_reader` plugin when trying to run it from Docker. I noticed that the current logic is specified as follows:

```python
    for cmd in ALLOWED_CMDS:
        if len(call(cmd + " --help")):
            logger.debug('AsciiDocReader: Using cmd: %s', cmd)
            return cmd
```

`len(call(cmd + " --help"))` seems to be non-zero i.e `True` even when checking for `asciidoc` which is not present in my Docker image.

This later causes the following error:

```
[21:13:38] WARNING  AsciiDocReader: strderr: b'/bin/sh: asciidoc: not  log.py:89
                    found\n'
```

PR #1402, opened 2024-06-05
[asciidoc_reader] use `shutil.which` to detect available binaries

tpwo:
`shutil.which` is a standard and multiplatform solution to perform checks like this.

This solves the issue described in #1401.


justinmayer (approved):
Thank you for the enhancement, Tomasz 🎖️

PR #1408, opened 2025-02-07
[representative_image] Also process draft & hidden articles & pages

Lucas-C:
I have a few hidden pages on a my website, and I could not figure why the representative image was incorrect.

This very short PR simply aims to generate representative images for ALL kind of articles & pages.

I tested it on my own blog, and it works fine.


justinmayer (approved):
Thanks for the enhancement, Lucas! 👍

PR #1410, opened 2025-03-11
Point `pelican_youtube` to repository HEAD (26-Jun-2021)

bittner:
The `pelican_youtube` submodule still points to a commit from 9-Dec-2017 while the repository HEAD is from 26-Jun-2021.

## References

- https://github.com/kura/pelican_youtube/commits/main/


justinmayer:
Thanks, Peter!

